PR #16, opened 2022-12-12
feat(cli): converts cli to be project driven making all commands operate in current dir

Encord-davids:


Encord-davids:
> Two considerations:
>
> 1. It seems redundant to add the `--target` parameter for basically every command, should we perhaps move it one level up? I am sure you thought of this so just wanted to hear.
> 2. When you run `encord-active visualise` outside a project directory, it opens the app but doesn't really show anything. I think that we should handle this better.
>
> Both things are something that we could potentially do in follow up PRs?

As for 1, I have considered it but I hate the fact all the usage would change to `encord-active --target="foo" download` which is counterintuitive.

And for 2, I haven't noticed it so I'll fix it before merging, thanks for catching this 💪

frederik-encord (approved):

PR #23, opened 2022-12-14
fix(model quality): Simplify performance by metrics

frederik-encord:


frederik-encord:
One screenshot will need to be updated in the docs, but it is already touched in #21 so to avoid conflicts and potentially errors by the wrong order of merges, I will leave this for a new PR.
Added issue #24.

Encord-davids (dismissed):

frederik-encord:
@Encord-davids Would you mind giving this another ✅

Encord-davids (approved):

PR #26, opened 2022-12-16
fix: don't predict grid defaults by screen size

Encord-davids:


src/encord_active/app/model_quality/sub_pages/__init__.py
```diff
@@ -73,49 +73,6 @@ def metric_details_description(metric_name: str = ""):
             st.markdown(f"### The {metric_meta['title']} metric")
             st.markdown(metric_meta["long_description"])
 
-    @staticmethod
-    def row_col_settings_in_sidebar():
```
frederik-encord, line 77: If we remove this, don't we need to remove the calls from, e.g., `src/encord_active/app/model_quality/sub_pages/false_positives.py`?

Encord-davids: the `super` class has this method defined already, this was just an override 🙂

frederik-encord: I see.

frederik-encord (approved):

PR #32, opened 2022-12-19
refactor: unifies db queries for state population

Encord-davids:


src/encord_active/app/common/state.py
```diff
@@ -68,3 +71,11 @@ def populate_session_state():
     st.session_state.data_dir = project_file_structure.data
     st.session_state.ontology_file = project_file_structure.ontology
     st.session_state.db_path = project_file_structure.db
+
+    DBConnection.set_project_path(st.session_state.project_dir)
```
frederik-encord, line 75: Do we need to set the path here and again in `set_page_config`? 
My intuition says that it should be enough to do it once?

Encord-davids: I believe I just moved it from `set_page_config` here

frederik-encord (approved):

PR #36, opened 2022-12-20
refactor: moves metrics loading and its deps to lib

Encord-davids:


frederik-encord (approved):

frederik-encord:
Just need to run isort because of the refactor.

PR #51, opened 2023-01-04
Fix (importing predictions): fix polygon points for segmentation

Gorkem-Encord:
Polygon points are converted to absolute values.


Encord-davids (dismissed):

Encord-davids:
Can we make this change ontop of our refactor branch? (otherwise, it might be lost when we merge)

PR #53, opened 2023-01-04
Fix import predictions

Gorkem-Encord:


Encord-davids:
This is a duplicate of this #54. Closing.

PR #54, opened 2023-01-04
Fix: validate imported polygon points

Gorkem-Encord:
Now `points_to_mask` checks and validates the input type.


src/encord_active/lib/model_predictions/writer.py
```diff
@@ -65,7 +65,12 @@ def polyobj_to_nparray(o: dict, width: int, height: int) -> np.ndarray:
 
 def points_to_mask(points: np.ndarray, width: int, height: int):
     mask = np.zeros((height, width), dtype=np.uint8)
-    mask = cv2.fillPoly(mask, [(points * np.array([[width, height]])).astype(int)], 1)  # type: ignore
+
+    if np.issubdtype(points.dtype, np.float):
+        if not (np.min(points) >= 0.0 and np.max(points) <= 1.0):
+            raise ValueError("Float polygon points between 0 and 1.")
+        points = points * np.array([[width, height]])
```
frederik-encord, line 72:
```suggestion
```

frederik-encord (changes requested):
Let's avoid raising the error. Could something like the suggested work? 
So basically if float and values <= 1, upscale.

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -65,7 +65,12 @@ def polyobj_to_nparray(o: dict, width: int, height: int) -> np.ndarray:
 
 def points_to_mask(points: np.ndarray, width: int, height: int):
     mask = np.zeros((height, width), dtype=np.uint8)
-    mask = cv2.fillPoly(mask, [(points * np.array([[width, height]])).astype(int)], 1)  # type: ignore
+
+    if np.issubdtype(points.dtype, np.float):
```
frederik-encord, line 69:
```suggestion
    if np.issubdtype(points.dtype, np.floating):
```
Should fix the mypy error

frederik-encord (changes requested):

PR #63, opened 2023-01-05
Refactor: enforce project file structure usage

eloy-encord:
**Context**: Latest addition of `ProjectFileStructure` class allows to declare paths via file system references in `ProjectFileStructure` instead of explicitly declaring how to construct such paths each time they are needed with the subsequent error/bug chances and loss of portability from one project file structure to another. At the time of writing this PR the explicit file structure version is predominant across the codebase and we aim to encapsulate usage of every file system related path through `ProjectFileStructure` class.

This PR starts enforcing `ProjectFileStructure` as the main and only way to access file/dir paths of project's elements (e.g. image, label row json, metadata). The following changes were applied:

- `project_dir` attribute was removed from `Project` class to ensure that every required path related content is accessed via `ProjectFileStructure`.
- Convert explicit paths of project's elements into their `ProjectFileStructure` versions.
- Misc: update `Project` related functions to work with `ProjectFileStructure` and convert their calls from outside the `Project` class to their `Project` instance method call equivalent.
- Misc: update import settings to avoid naming redundancies like `from encord_active.lib.project.project import Project` (after this PR it will look like `from encord_active.lib.project import Project`)

### DIFF
--- a/src/encord_active/app/common/state.py
+++ b/src/encord_active/app/common/state.py
@@ -10,7 +10,7 @@ from encord_active.lib.db.merged_metrics import MergedMetrics
 from encord_active.lib.db.tags import Tag, Tags
 from encord_active.lib.metrics.utils import MetricData
 from encord_active.lib.model_predictions.reader import LabelSchema, OntologyObjectJSON
-from encord_active.lib.project.project_file_structure import ProjectFileStructure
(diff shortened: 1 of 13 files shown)


Encord-davids (approved):
💪

PR #71, opened 2023-01-09
fix: metric importance catch value error

frederik-encord:


src/encord_active/app/model_quality/sub_pages/metrics.py
```diff
@@ -88,12 +88,19 @@ def build(
...
-            st.altair_chart(chart, use_container_width=True)
+                try:
+                    chart = create_metric_importance_charts(
+                        model_predictions,
+                        metric_columns=metric_columns,
+                        num_samples=num_samples,
+                    )
+                    st.altair_chart(chart, use_container_width=True)
+                except ValueError as e:
+                    __import__("ipdb").set_trace()
```
Encord-davids, line 99: probably without this right? 🙈

frederik-encord: Stupid me. I should add an automatic check for that! Fixed

Encord-davids (approved):

PR #73, opened 2023-01-09
Chore (example): maskrcnn example

Gorkem-Encord:
This is an end-to-end project that can be used to train a mask-RCNN model.
- Utilizes maskRCNN model in the Torchvision package.
- Main training pipeline is implemented with PyTorch.
- Uses wandb platform to log metrics.
- Encord-Active compatible predictions can be generated.
- Utility scripts to downscale dataset or visualizing prediction results are added.
- User will only need to provide a config.ini file for the training and generating predictions


.gitignore
```diff
@@ -352,3 +352,4 @@ Sessionx.vim
 
 imports.prof
 ci-tests
+examples/maskrcnn-example/config.ini
```
frederik-encord, line 355: I think that this should go to a separate `.gitignore` file in the `maskrcnn-example` directory.

examples/maskrcnn-example/example_config.ini
```diff
@@ -0,0 +1,32 @@
...
+
+[INFERENCE]
+# Fill this part after the training
+# Check wandb_id on the wandb platform according to experiment
+target_data_folder = /path/to/target/folder
+target_ann = /path/to/target/annotation/file
+wandb_id = wandb_unique_id
+model_checkpoint_path = /path/to/model/checkpoint
+ontology_filepath = /path/to/ontology.json
+confidence_threshold = 0.5
```
frederik-encord, line 32: This section is not really self-explanatory. Could we add some additional comments there?

frederik-encord (changes requested):

.gitignore
Gorkem-Encord: removed this from main .gitignore, created a new one for the masrcnn-example, added wandb folder and config.ini file.

frederik-encord (approved):

PR #77, opened 2023-01-10
feat(examples): notebooks for the example directory

frederik-encord:
fix(examples): get rid of ngrok

feat(examples): readme for the examples directory

feat(examples): build your own custom metric


examples/download-sandbox-dataset.ipynb
```diff
@@ -0,0 +1,92 @@
...
+    "project_name = PREBUILT_PROJECT_TO_STORAGE[int(input('Project no:'))]\n",
+    "\n",
+    "# Choose where to store the data\n",
+    "project_path = Path.cwd() / project_name\n",
+    "\n",
+    "# Download the dataset\n",
+    "fetch_prebuilt_project(project_name, project_path)\n",
+    "\n",
+    "# Open the app\n",
+    "!encord-active visualise -t $project_path"
```
Gorkem-Encord, line 63: if the project_path includes special characters (like '[' for example), it cannot open, so it should be between quotation

Gorkem-Encord:
in getting-started-with-encord-projects.ipynb
image

Gorkem-Encord:
in getting-started-with-coco-project.ipynb, since we already have the following commands in the later cells, I think we can just remove them, otherwise they will be just repeating information

image

Gorkem-Encord:
In building-a-custom-metric-function.ipynb, since `data_unit` example structure is not complete (with other types of objects and classifications), we can mention that _you can inspect the full label-row in .../data/<label_hash>/label\_row.json_

Gorkem-Encord:
In building-a-custom-metric-function.ipynb, in the example metric class (cell 7), both image-level metric and object-level metric is written to the same file (they are using the same writer), are we okay with this? If not, we should warn users that they can only write either image-level score or object-level score in a metric class, otherwise this example may lead to errors.

Gorkem-Encord:
Overall, very well and clean notebook files!

frederik-encord:
@Gorkem-Encord I have fixed your comments.

Gorkem-Encord (approved):

PR #86, opened 2023-01-12
feat(classification): adds classification answers as class labels

Encord-davids:


src/encord_active/lib/metrics/semantic/img_classification_quality.py
```diff
@@ -56,26 +70,33 @@ def __init__(self, num_nearest_neighbors: int = 10, certainty_ratio: float = 0.6
...
+            answer=self.index_to_answer_name[question_hash][gt_label],
+            target_answer=self.index_to_answer_name[question_hash][target_label],
+            above_threshold=target_label_frequency > threshold,
+        )
 
-        else:  # covers cases for  target_label_frequency <= threshold:
-            description = f":question: For the question `{self.featureNodeHash_to_question_name[question_hash]}`, the image is annotated as `{self.index_to_answer_name[question_hash][gt_label]}`. \
+    def build_description_string(self, info: DescriptionInfo) -> str:
+        if not info["above_threshold"]:
+            return f":question: For the question `{info['question']}`, the image is annotated as `{info['question']}`. \
```
frederik-encord, line 92:
```suggestion
            return f":question: For the question `{info['question']}`, the image is annotated as `{info['answer']}`. \
```

frederik-encord (changes requested):
One little bug. Otherwise looks good.

src/encord_active/lib/metrics/semantic/img_classification_quality.py
Encord-davids: Good catch!

frederik-encord (approved):

PR #89, opened 2023-01-16
Fix (maskrcnn-example): improvements on maskrcnn-example

Gorkem-Encord:
- Bounding boxes that have non-positive width/height raises errors during the training. We skip those annotations by informing user which image has that annotation.
- Readme improved
- Workflow for fixing labels added.


eloy-encord:
This PR should point to #88 instead of `main`, right?

Gorkem-Encord:
Since #88 is merged, It points to the main now @eloy-encord

examples/maskrcnn-example/utils/encord_dataset.py
```diff
@@ -21,6 +21,16 @@ def __getitem__(self, idx):
 
         boxes, labels, area, iscrowd = [], [], [], []
         for target_item in target:
+            if (not target_item["bbox"][2] > 0) or (not target_item["bbox"][3] > 0):
```
eloy-encord, line 24:
```suggestion
            if target_item["bbox"][2] <= 0 or target_item["bbox"][3] <= 0:
```

examples/maskrcnn-example/utils/encord_dataset.py
```diff
@@ -21,6 +21,16 @@ def __getitem__(self, idx):
...
+                print(
```
eloy-encord, line 25: 💭 probably best if added as a log to be able to silence it

Gorkem-Encord: I think we should also warn users that there are annotations like that so that they can fix/remove those. What do you think is the best way to give a message to the user?

Gorkem-Encord: @eloy-encord

eloy-encord: I would go with logs as we have done before in EA, or an optional `verbose` argument that if set to `True` it will show the warnings.

eloy-encord:
for some reason, now this PR contains a lot more of content supposedly not related to the PR, what is the current approach with this PR? all in one PR and modify its name to reflect the actual content or perhaps a mistake with the branchs?

Gorkem-Encord:
@eloy-encord I merged another PR pointing to this branch

eloy-encord (approved):

PR #97, opened 2023-01-18
 fix: use default ssh key path when importing local project

yogesh-encord:


src/encord_active/lib/coco/importer.py
```diff
@@ -14,6 +14,7 @@
 from PIL import ImageOps
 from tqdm.auto import tqdm
 
+from encord_active.cli.config import app_config
```
Encord-davids, line 17: A general comment on imports:
Our app is structured into 3 projects. In the `src/encord_active` folder we have 3 modules, `app` `lib` and `cli`. We structured it this way so we could eventually split it into separate redistributable. In order to allow such thing in the future, we should scope our import to the eventual dependencies between the projects.

`app` depends on `lib`
`cli` depends on `lib` and `app`
`lib` doesn't have any dependencies.

I would suggest getting the ssh key as an argument.

yogesh-encord: That makes sense, removed the offending import

src/encord_active/cli/utils/coco.py
```diff
@@ -69,6 +70,7 @@ def import_coco_project(images_dir: Path, annotations_file: Path, target: Path,
     coco_importer.create_project(
         dataset=dataset,
         ontology=ontology,
+        ssh_key_path=app_config.get_or_query_ssh_key(),
```
Encord-davids, line 73: Now that I think of this, we shouldn't really force the user to add and ssh key if they only intend to use the product locally.

Can we replace this add the key only if set in the config without prompting the user if it isn't set?

yogesh-encord: Replaced with just a getter!

Encord-davids (dismissed):

Encord-davids (approved):

PR #98, opened 2023-01-18
feat(import): init project from data directory

frederik-encord:
The functionality for importing a local dataset based off glob.

Documentation will come in a separate PR shortly.

### DIFF
--- a/src/encord_active/cli/main.py
+++ b/src/encord_active/cli/main.py
@@ -3,6 +3,7 @@ from typing import List, Set, TypedDict
 
 import rich
 import typer
+from rich.markup import escape
 from rich.panel import Panel
 
 import encord_active.app.conf  # pylint: disable=unused-import
@@ -11,7 +12,6 @@ from encord_active.cli.imports import import_cli
 from encord_active.cli.print import print_cli
 from encord_active.cli.utils.decorators import bypass_streamlit_question, ensure_project
 from encord_active.cli.utils.prints import success_with_visualise_command
-from encord_active.lib.project.local import NoFilesFoundError, ProjectExistsError
 
 cli = typer.Typer(
     rich_markup_mode="rich",
@@ -88,7 +88,7 @@ def import_local_project(
         file_okay=False,
     ),
     glob: List[str] = typer.Option(
-        ["**/*.jpg", "**/*.png", "**/*.tiff"],
+        ["**/*.jpg", "**/*.png", "**/*.jpeg", "**/*.tiff"],
         "--glob",
         "-g",
         help='Glob pattern to choose files in the "leaf directories". Note that you can repeat the `--glob` argument if you want to match multiple things.',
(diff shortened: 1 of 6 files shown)


src/encord_active/cli/main.py
```diff
@@ -73,23 +75,105 @@ def download(
...
-            style="green",
-            expand=False,
+    try:
+        project_path = init_local_project(
+            root=root, target=target, glob=glob, project_name=project_name, symlinks=symlinks, dryrun=dryrun
         )
-    )
+        if dryrun:
+            if not isinstance(project_path, list):
+                raise ValueError("Expected a list of paths for the dryrun execution")
```
Encord-davids, line 132: [curious]🤔

Why are we expecting a list if its a dry run but a path if not?

src/encord_active/cli/main.py
```diff
@@ -73,23 +75,105 @@ def download(
...
+                expand=False,
+                style="yellow",
+            )
+        )
+        typer.Abort()
+    except ProjectExistsError as e:
+        rich.print(
+            Panel(
+                f"""
+{e.args[0]}
```
Encord-davids, line 167: [nitpick] 💅

I think it's cleaner 

```suggestion
{str(e)}
```

src/encord_active/lib/encord/local_sdk.py
```diff
@@ -53,42 +56,50 @@ class LabelRowMetadata(TypedDict):
...
+
 class LocalDataRow(DataRow):
     def __init__(
         self,
         uid: str,
         label_hash: str,
         title: str,
         data_type: DataType,
-        path: Path,
+        frame: Union[DataRowFrame, List[DataRowFrame]],
```
Encord-davids, line 75: [curious]🤔 

Should we maybe name this differently since we are already using frame differently in other contexts (frame mostly refers to frame number).

of the top of my head is something like `media` maybe?

src/encord_active/cli/main.py
frederik-encord: This is because if we do a dryrun, we want to print all the stuff, but I didn't want the printing to be in the lib. I wanted that to be in the CLI. So in order to be able to print int, I had to return the list of files.

Perhaps there is a better way to do this? 🤷🏻

src/encord_active/lib/project/local.py
```diff
@@ -0,0 +1,125 @@
...
+
+    if glob is None:
+        glob = ["**/*.jpg", "**/*.jpeg", "**/*.png", "**/*.tiff"]
+
+    files = list(chain(*[root.glob(g) for g in glob]))
+    if dryrun:
+        return files
+
+    if not len(files):
+        raise NoFilesFoundError()
```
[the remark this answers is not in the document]
frederik-encord: If people want to use the `init_local_project` via code, I thought that it would be nicer to return an empty list with the dryrun than raising an exception?
At least that would make their code prettier.
But we can swap the order.

src/encord_active/lib/encord/local_sdk.py
frederik-encord: media is nice because it is the same for singular and plural. Fixed it.

src/encord_active/cli/main.py
Encord-davids: Ok, I understand the reasoning now. Printing in the lib isn't what we want but we do want consistency between dry runs and regular runs.

There might be a better way to decouple this. If we move the parts that are writing to the disk into a separate `save` operation (another method/return a callback/create a class that wraps the behaviour) we can avoid some of the complexity and we won't need to pass the `dryrun` outside of the cli.

Let me know what you think, I realise this might not be the smallest change, and probably not critical for now.

src/encord_active/lib/project/local.py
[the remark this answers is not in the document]
Encord-davids: I personally hate it when I run something with a dry run flag and it isn't consistent with real runs. But I don't think it is a critical inconsistency, so up to you to decide how to go about this 🙂

src/encord_active/lib/project/local.py
```diff
@@ -0,0 +1,131 @@
...
+    based on the `glob` arguments.
+
+    Args:
+        root: The root from which to search for files.
+        target: The directory in which the new project will be stored.
+        glob: A list of glob expressions that will be used to match image files.
+        project_name: A specific project name to use. If left out, the project
+            name will be the root directory name prepended with "[EA] "
+        symlinks: If false, files will be copied. If true, files will be symlinked.
+        dryrun: If false, a project will be created. If false, a list of all the
```
eloy-encord, line 53:
```suggestion
        dryrun: If false, a project will be created. If true, a list of all the
```

Encord-davids: there isn't a dry run anymore so this line can be deleted 🙂

Encord-davids (approved):

Encord-davids (approved):

PR #100, opened 2023-01-19
Feat: new data quality summary dashboard

Gorkem-Encord:
- I tried to highlight the main properties of the dataset
- Severe and moderate outliers are visualized in a single plot
- Image size distribution is added

One thing I am not happy with is the 'Issues' pane on the right side. It can be more user-friendly, however, my HTML/CSS knowledge limits me here. Another problem with this section is that if there are too many issues, it will make that pane quite long, so they should be scrollable, but I am not sure how to do that with native streamlit components.


src/encord_active/lib/dataset/outliers.py
```diff
@@ -6,8 +7,15 @@
 from encord_active.lib.metrics.utils import MetricSchema
 
 
+class OutlierStatus(str, Enum):
```
Encord-davids, line 10: [nitpick] 💅

This can just be `Severity` and we could reuse it to categorize other insights in the future (not just outliers)

```suggestion
class Severity(str, Enum):
```

src/encord_active/lib/dataset/outliers.py
```diff
@@ -44,6 +52,16 @@ def get_iqr_outliers(
...
 
+    df[_COLUMNS.outliers_status] = OutlierStatus.low.value
+    df.loc[
+        ((severe_lb <= df[_COLUMNS.score]) & (df[_COLUMNS.score] < moderate_lb))
+        | ((severe_ub >= df[_COLUMNS.score]) & (df[_COLUMNS.score] > moderate_ub)),
+        _COLUMNS.outliers_status,
+    ] = OutlierStatus.moderate.value
+    df.loc[
+        (df[_COLUMNS.score] < severe_lb) | (df[_COLUMNS.score] > severe_ub), _COLUMNS.outliers_status
+    ] = OutlierStatus.severe.value
```
Encord-davids, line 63: [nitpick] 💅

when using string enums there is not need to compare string values.

```suggestion
    df[_COLUMNS.outliers_status] = OutlierStatus.low
    df.loc[
        ((severe_lb <= df[_COLUMNS.score]) & (df[_COLUMNS.score] < moderate_lb))
        | ((severe_ub >= df[_COLUMNS.score]) & (df[_COLUMNS.score] > moderate_ub)),
        _COLUMNS.outliers_status,
    ] = OutlierStatus.moderate
    df.loc[
        (df[_COLUMNS.score] < severe_lb) | (df[_COLUMNS.score] > severe_ub), _COLUMNS.outliers_status
    ] = OutlierStatus.severe
```

src/encord_active/app/data_quality/sub_pages/summary.py
```diff
@@ -1,29 +1,148 @@
...
+                        }
+                    ),
+                ],
+                axis=0,
+            )
+
+            for _, row in df.iterrows():
+                if row[_COLUMNS.outliers_status] == OutlierStatus.severe.value:
+                    total_severe_outliers.add(row[_COLUMNS.identifier])
+                elif row[_COLUMNS.outliers_status] == OutlierStatus.moderate.value:
```
Encord-davids, line 88: [nitpick] 💅

when using string enums there is not need to compare string values.

```suggestion
                if row[_COLUMNS.outliers_status] == OutlierStatus.severe:
                    total_severe_outliers.add(row[_COLUMNS.identifier])
                elif row[_COLUMNS.outliers_status] == OutlierStatus.moderate:
```

Encord-davids, line 88: [nitpick] 💅

when using string enums there is not need to compare string values.

```suggestion
                if row[_COLUMNS.outliers_status] == OutlierStatus.severe:
                    total_severe_outliers.add(row[_COLUMNS.identifier])
                elif row[_COLUMNS.outliers_status] == OutlierStatus.moderate:
```

src/encord_active/app/common/components/data_quality_summary.py
```diff
@@ -0,0 +1,23 @@
+from typing import Union
+
+import streamlit as st
+
+
+def summary_item(
+    section: st._DeltaGenerator,
```
Encord-davids, line 7: to increase reusability, should we remove the section and just return a string or an `st.markdown`?

(in the current way I would be able to render this item without a container)

Encord-davids:
When I run this locally, the page renders are a bit too slow IMO. I didn't get a chance to deep dive for the reason but may I suggest we spend a little more time before releasing this to make the experience a little snappier.

src/encord_active/app/common/components/data_quality_summary.py
Gorkem-Encord: Ok. Now, summary_item only returns html string.

Encord-davids (approved):

Encord-davids (approved):

PR #101, opened 2023-01-20
fix(predictions): properly draw bounding box predictions

frederik-encord:
Bounding box predictions doesn't work for two reasons:

1. The dataframe schemas that we defined for the predictions didn't allow empty RLEs. When a bounding box prediction is imported in the app, the rle will be NaN in the data frames because they don't have a mask. This is fixed.
2. The plotting of the predictions in the TP, FP, and FN pages were missing objects, as they expected an RLE. This is fixed.

We should introduce an object detection dataset to the sandbox datasets to be able to test these things regularly.

### DIFF
--- a/src/encord_active/lib/common/image_utils.py
+++ b/src/encord_active/lib/common/image_utils.py
@@ -43,22 +43,23 @@ def draw_object(
     hex_color = color.value if isinstance(color, Color) else color
     _color: Tuple[int, ...] = hex_to_rgb(hex_color)
     _color_outline: Tuple[int, ...] = hex_to_rgb(hex_color, lighten=-0.5)
-    if isinstance(row["rle"], str):
-        if with_box:
-            box = get_bbox_csv(row)
-            image = cv2.polylines(image, [box], isClosed, _color, thickness // 2, lineType=cv2.LINE_8)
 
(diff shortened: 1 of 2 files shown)


src/encord_active/lib/common/image_utils.py
```diff
@@ -43,11 +43,13 @@ def draw_object(
...
-        if with_box:
-            box = get_bbox_csv(row)
-            image = cv2.polylines(image, [box], isClosed, _color, thickness // 2, lineType=cv2.LINE_8)
 
+    has_mask = isinstance(row["rle"], str)
+    if with_box or not has_mask:
+        box = get_bbox_csv(row)
+        image = cv2.polylines(image, [box], isClosed, _color, thickness, lineType=cv2.LINE_8)
+
+    if has_mask:
```
Encord-davids, line 52: [nitpick] 💅 

An early return makes it read slightly better

```suggestion
    if with_box or not has_mask:
        box = get_bbox_csv(row)
        image = cv2.polylines(image, [box], isClosed, _color, thickness, lineType=cv2.LINE_8)
        return image

```

frederik-encord: @Encord-davids fixed this

Encord-davids (approved):

PR #105, opened 2023-01-20
fix: ensure embeddings folder exists

Encord-davids:


src/encord_active/lib/embeddings/cnn.py
```diff
@@ -292,14 +289,18 @@ def get_cnn_embeddings(
...
+        cnn_embeddings = generate_cnn_object_embeddings(iterator)
     elif embedding_type == EmbeddingType.CLASSIFICATION:
-        generate_cnn_classification_embeddings(iterator, filepath=target)
+        cnn_embeddings = generate_cnn_classification_embeddings(iterator)
+    else:
+        raise ValueError(f"Unsupported embedding type {embedding_type}")
+
+    target_path = Path(target)
+    target_path.parent.mkdir(parents=True, exist_ok=True)
+    target_path.write_bytes(pickle.dumps(cnn_embeddings))
```
Gorkem-Encord, line 302: Do you think shall we encapsulate here within try-finally block, so that if there is an error in embedding extraction an empty file will still be guaranteed to be written to a file?

Encord-davids: Would that be the desired behaviour? I don't think we should write empty files, if no file is there it is clear it needs to be generated.

Gorkem-Encord (approved):

PR #110, opened 2023-01-23
Refactor: summary dashboards

Gorkem-Encord:
Data quality summary and label quality summary pages have different information to report; therefore, their layout will be different in the future. They need to be separated.

### DIFF
--- a/src/encord_active/app/common/components/metric_summary.py
+++ b/src/encord_active/app/common/components/metric_summary.py
@@ -16,76 +16,22 @@ from encord_active.lib.charts.data_quality_summary import (
 from encord_active.lib.common.image_utils import show_image_and_draw_polygons
 from encord_active.lib.dataset.outliers import (
     IqrOutliers,
-    MetricOutlierInfo,
-    MetricsSeverity,
     MetricWithDistanceSchema,
-    Severity,
-    get_iqr_outliers,
+    get_all_metrics_outliers,
 )
 from encord_active.lib.dataset.summary_utils import (
     get_all_annotation_numbers,
     get_all_image_sizes,
-    get_median_value_of_2D_array,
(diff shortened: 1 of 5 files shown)


src/encord_active/app/common/components/metric_summary.py
```diff
@@ -1,22 +1,113 @@
...
 import pandas as pd
 import streamlit as st
 from pandera.typing import DataFrame
+from streamlit.delta_generator import DeltaGenerator
 
 from encord_active.app.common.components import build_data_tags
 from encord_active.app.common.components.data_quality_summary import summary_item
 from encord_active.app.common.components.tags.individual_tagging import multiselect_tag
-from encord_active.app.common.state import get_state
+from encord_active.app.common.state import MetricOutlierInfo, MetricsSeverity, get_state
```
frederik-encord, line 11: Should `MetricOutlierInfo` and `MetricsSeverity` really belong to the state class? I would probably put them somewhere else. Probably in `lib/dataset/outliers.py`.

src/encord_active/app/common/components/metric_summary.py
```diff
@@ -1,22 +1,113 @@
...
+                        "total_moderate_outliers": [item.iqr_outliers.n_moderate_outliers],
+                    }
+                ),
+            ],
+            axis=0,
+        )
+
+    all_metrics_outliers.sort_values(by=["total_severe_outliers"], ascending=False, inplace=True)
+
+    return all_metrics_outliers
```
frederik-encord, line 86: These should probably go to the lib as well. It doesn't depend on streamlit so it can be reused without the frontend.

frederik-encord (changes requested):

PR #113, opened 2023-01-23
fix(embedding): check for the correct list for matching embedding

Encord-davids:


yogesh-encord (approved):

src/encord_active/lib/embeddings/cnn.py
```diff
@@ -198,7 +198,7 @@ def generate_cnn_classification_embeddings(iterator: Iterator) -> List[LabelEmbe
             and collection["frame"] == iterator.frame
         ]
 
-        if not len(image_collections):
+        if not len(matching_image_collections):
```
yogesh-encord, line 201: Could even just do `if not matching_image_collections` right?

Encord-davids: True 👍

yogesh-encord (approved):

PR #121, opened 2023-01-25
feat(import): adds re-import behaviour

Encord-davids:


src/encord_active/cli/imports.py
```diff
@@ -63,7 +66,7 @@ def import_project(
     # Encord
     encord_project_hash: Optional[str] = typer.Option(
         None,
-        "--project_hash",
+        "--project-hash",
```
frederik-encord, line 69: Nice! Have wanted to do this for a while.

frederik-encord (approved):

frederik-encord (approved):

PR #122, opened 2023-01-25
Feat (dashboard): class distribution plots

Gorkem-Encord:
Class distribution plots are added to the label summary dashboards
- Object types (bounding boxes, polygons etc. ) is visualized on a single plot
- Classification types are visualized separately. In other words, each classification question has its own plot.
- Under sampled values are marked as red.


src/encord_active/lib/dataset/summary_utils.py
```diff
@@ -30,19 +30,62 @@ def get_median_value_of_2D_array(array: np.ndarray) -> np.ndarray:
...
+                        ][0]
+                        classification_question_name = classification_answer_item["name"]
+                        if classification_question_name in labels["classifications"]:
+
+                            if isinstance(classification_answer_item["answers"], list):
+                                for answer_item in classification_answer_item["answers"]:
+                                    if answer_item["name"] in labels["classifications"][classification_question_name]:
+                                        labels["classifications"][classification_question_name][
+                                            answer_item["name"]
+                                        ] += 1
```
Encord-davids, line 79: if this is for handling checkboxes, this is probably not needed

Encord-davids (changes requested):

src/encord_active/lib/dataset/summary_utils.py
```diff
@@ -30,19 +30,62 @@ def get_median_value_of_2D_array(array: np.ndarray) -> np.ndarray:
...
     classification_label_counter = 0
     object_label_counter = 0
 
+    project_ontology = json.loads((project_folder / "ontology.json").read_text(encoding="utf-8"))
+    for object_item in project_ontology["objects"]:
+        labels["objects"][object_item["name"]] = 0
+    for classification_item in project_ontology["classifications"]:
+        labels["classifications"][classification_item["attributes"][0]["name"]] = {}
+
+        # For radio and checkbox types
```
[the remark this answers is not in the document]
Gorkem-Encord: I think presenting a comprehensive view of the labels in the dashboard is better. This part is just for reporting and does not affect the other parts, so we should also report checklist and text type classifications as our customers are also using those.

src/encord_active/lib/dataset/summary_utils.py
Gorkem-Encord: This is both for the radio and checklist types.

src/encord_active/lib/dataset/summary_utils.py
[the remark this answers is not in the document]
Encord-davids: They are using those but won't be able to find them anywhere else in the app, which might be a bit confusing.

src/encord_active/lib/dataset/summary_utils.py
```diff
@@ -41,22 +53,67 @@ def get_median_value_of_2d_array(array: np.ndarray) -> np.ndarray:
...
     object_label_counter = 0
 
-    for label_row in (project_folder / "data").iterdir():
+    project_ontology = json.loads((project_paths.ontology).read_text(encoding="utf-8"))
+    ontology = OntologyStructure.from_dict(project_ontology)
+
+    for object_item in ontology.objects:
+        labels.objects[object_item.name] = 0
+    for classification_item in ontology.classifications:
+        labels.classifications[classification_item.attributes[0].name] = {}
```
Encord-davids, line 72: Probably can be replaced with `setdefault` since we already have a case where we use it.

src/encord_active/lib/dataset/summary_utils.py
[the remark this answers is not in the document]
Gorkem-Encord: I think it might be better if we present them otherwise there will be missing information and an incomplete view of the dataset/labels, but you might be right in terms of other places. Let's ask @frederik-encord for the final decision. Should we include class distribution charts for the checklist and text-type classifications?

src/encord_active/lib/dataset/summary_utils.py
Gorkem-Encord: Do you mean like the following:

```suggestion
    for object_item in ontology.objects:
        labels.objects.setdefault(object_item.name, 0)
    for classification_item in ontology.classifications:
        labels.classifications.setdefault(classification_item.attributes[0].name, {})
```

Gorkem-Encord: @Encord-davids

Encord-davids: Sorry I wasn't clear, I meant in the main loop where we increase the counters.

Gorkem-Encord: If I understood correctly, we need to initialize all of them from zero because, in the label_row, there might not be labels for all object/classification types, so in this case we need to report that there is a class X in your ontology but it has no labels. If we do it only iterating through label_rows we cannot know it.

src/encord_active/lib/dataset/summary_utils.py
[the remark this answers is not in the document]
Gorkem-Encord: @frederik-encord

src/encord_active/lib/dataset/summary_utils.py
Encord-davids: Ahh totally makes sense, sorry for the oversight from my side 🙇

src/encord_active/lib/dataset/summary_utils.py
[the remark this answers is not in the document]
frederik-encord: I think that we should include them for the following reason. If users see their checkboxes in the summary dashboard but not elsewhere, hopefully, they will reach out to us and "pull" the remaining features around checkboxes out of us.

Encord-davids (approved):

PR #124, opened 2023-01-26
feat(app): show project title in sidebar

frederik-encord:


src/encord_active/app/streamlit_entrypoint.py
```diff
@@ -68,6 +69,8 @@ def main(project_path: str):
     State.init(project_dir)
 
     with st.sidebar:
+        project_meta = fetch_project_meta(project_dir)
+        st.markdown(f"`project:` {project_meta.get('project_title', project_dir.name)}")
```
Encord-davids, line 73: I think this should be enough no?

```suggestion
        st.header(project_meta.get("project_title", project_dir.name))
```

frederik-encord: Header is too big. I ended up using `st.subheader` instead 👍🏻

Encord-davids (approved):

PR #127, opened 2023-01-26
Feat (dashboard): 2D metric plots

Gorkem-Encord:
In dashboard, 2 different metrics can be visualized wrt each other in a scatter plot.
The layout can be arranged in another PR as it requires implementing custom frontend components.


src/encord_active/lib/charts/data_quality_summary.py
```diff
@@ -94,3 +94,24 @@ def create_image_size_distribution_chart(image_sizes: np.ndarray) -> go.Figure:
     )
 
     return fig
+
+
+def create_2d_metric_chart(metrics_df: pd.DataFrame, show_trendline: bool = True) -> go.Figure:
```
frederik-encord, line 99: Can we use a `pandera` schema here?

src/encord_active/lib/charts/data_quality_summary.py
```diff
@@ -94,3 +94,24 @@ def create_image_size_distribution_chart(image_sizes: np.ndarray) -> go.Figure:
...
+    """
+    This function expects a dataframe where its columns are [identifier, x_value, y_value]
+    """
+    fig = px.scatter(
+        metrics_df,
+        x=metrics_df.columns[1],
+        y=metrics_df.columns[2],
```
frederik-encord, line 106: These are also very unsafe. Use the `pandera` schema to index columns instead.

src/encord_active/lib/charts/data_quality_summary.py
```diff
@@ -94,3 +94,24 @@ def create_image_size_distribution_chart(image_sizes: np.ndarray) -> go.Figure:
...
+        hover_data=["identifier"],
+        trendline="ols" if show_trendline else None,
+        trendline_color_override="black",
+    )
+    fig.update_layout(
+        title=f"{metrics_df.columns[1]} vs. {metrics_df.columns[2]}",
+        xaxis_title=metrics_df.columns[1],
+        yaxis_title=metrics_df.columns[2],
```
frederik-encord, line 114: Same thing.

src/encord_active/lib/dataset/outliers.py
```diff
@@ -89,7 +89,7 @@ def get_iqr_outliers(
 
 def get_all_metrics_outliers(metrics_data_summary: MetricsSeverity) -> pd.DataFrame:
```
frederik-encord, line 90: Schema for this return type would be great.

src/encord_active/app/common/components/metric_summary.py
```diff
@@ -28,13 +30,63 @@
...
 
 _COLUMNS = MetricWithDistanceSchema
 
 
+def render_2d_metric_plots(metrics_data_summary: MetricsSeverity):
+    with st.expander("2D metrics view", True):
+        metric_selection_col, scatter_plot_col = st.columns([2, 5])
+        metric_names = [metric_name for metric_name in metrics_data_summary.metrics.keys()]
+        x_metric_name = metric_selection_col.selectbox("x axis", metric_names, index=0)
+        y_metric_name = metric_selection_col.selectbox("y axis", metric_names, index=1)
```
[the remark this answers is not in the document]
frederik-encord: I mean taking index 0 and 1 for an empty `metric_names` list.

src/encord_active/app/common/components/metric_summary.py
```diff
@@ -28,13 +30,63 @@
...
+        trend_selected = metric_selection_col.checkbox(
+            "Show trend",
+            value=True,
+            help="Draws a trend line to demonstrate the relationship between the two metrics.",
+        )
+
```
frederik-encord, line 52: We use the `MetricSchema.[whatever]` in many places here. Perhaps we should type the `MetricOutlierInfo.df` as a `MetricSchema`?

src/encord_active/app/common/components/metric_summary.py
```diff
@@ -28,13 +30,63 @@
...
 )
 
 _COLUMNS = MetricWithDistanceSchema
 
 
+def render_2d_metric_plots(metrics_data_summary: MetricsSeverity):
+    with st.expander("2D metrics view", True):
+        metric_selection_col, scatter_plot_col = st.columns([2, 5])
+        metric_names = [metric_name for metric_name in metrics_data_summary.metrics.keys()]
+        x_metric_name = metric_selection_col.selectbox("x axis", metric_names, index=0)
```
frederik-encord, line 45: I would propose to use the `format_func` of the `selectbox` instead to directly get the MetricOutlierInfo class instead. Something like this (untested):

```suggestion
        x_outlier_info = metric_selection_col.selectbox("x axis", metrics_data_summary.metrics.values(), format_func=lambda m: m.name, index=0)
        x_metric_df = x_outlier_info.df[[MetricSchema.identifier, MetricSchema.score]]
```

This would make the code much cleaner.

frederik-encord: You can probably even get rid of the dictionary - and perhaps even sort the list on names to be very nice to the user :-)

src/encord_active/app/common/components/metric_summary.py
```diff
@@ -28,13 +30,63 @@
...
+        y_metric_df.rename(columns={MetricSchema.score: f"{y_metric_name}"}, inplace=True)
+
+        if len(x_metric_df.iloc[0][MetricSchema.identifier].split("_")) == len(
+            y_metric_df.iloc[0][MetricSchema.identifier].split("_")
+        ):
+            merged_metrics = pd.merge(x_metric_df, y_metric_df, how="inner", on=MetricSchema.identifier)
+        else:
+            x_changed, to_be_parsed_df = (
+                (True, x_metric_df.copy(deep=True))
+                if len(x_metric_df.iloc[0][MetricSchema.identifier].split("_")) == 4
```
frederik-encord, line 66: Are we sure that the dataframe is never empty?

src/encord_active/app/common/components/metric_summary.py
```diff
@@ -28,13 +30,63 @@
...
+                else (False, y_metric_df)
+            )
+
+            to_be_parsed_df[[MetricSchema.identifier, "identifier_rest"]] = to_be_parsed_df[
+                MetricSchema.identifier
+            ].str.rsplit("_", n=1, expand=True)
```
frederik-encord, line 72: It seems like `to_be_parsed_df` can both be `x_metric_df` and `y_metric_df` but we only check the length of the key of `x_metric_df`. Could this potentially go wrong if `y_metric_df` is an image-level metric?

frederik-encord (changes requested):

pyproject.toml
```diff
@@ -69,6 +69,7 @@ pandera = "^0.13.4"
 jupyterlab = {version = "^3.5.2", optional = true}
 ipywidgets = {version = "^8.0.4", optional = true}
 inquirerpy = "^0.3.4"
+statsmodels = "^0.13.5"
```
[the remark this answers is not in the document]
Gorkem-Encord: For the trendline feature of plotly, it is needed https://plotly.com/python/linear-fits/

_Plotly Express allows you to add [Ordinary Least Squares](https://en.wikipedia.org/wiki/Ordinary_least_squares) regression trendline to scatterplots with the trendline argument. In order to do so, you will need to [install statsmodels and its dependencies](https://www.statsmodels.org/stable/install.html)._

src/encord_active/app/common/components/metric_summary.py
Gorkem-Encord: I typed the `MetricOutlierInfo.df` as `MetricSchema`. However, I am not sure how that will relate to the usage of `MetricSchema.[whatever]` ?

src/encord_active/app/common/components/metric_summary.py
Gorkem-Encord: In this else block, either `x_metric_df` or `y_metric_df` has length of four when split, so if it is not x, it should be y automatically.

src/encord_active/app/common/components/metric_summary.py
Gorkem-Encord: I changed the type of `MetricOutlierInfo.df` to `MetricWithDistanceSchema` as it is more appropriate.

src/encord_active/app/common/components/metric_summary.py
Gorkem-Encord: I did this as you suggested; however, it raised an internal streamlit error that I could not solve, spent nearly 2 hours on that, so I reverted it back. Here is the error output when I use format_func in selectbox:

image

frederik-encord (approved):

PR #128, opened 2023-01-27
fix(import): local data row remove inheritance from encord sdk

frederik-encord:


frederik-encord:
@eloy-encord this should fix the issue that you had with the data row. I see that the encord SDK has changed a lot of times around the DataRow, so I removed the dependency to avoid this happening in the future.

Encord-davids (dismissed):

eloy-encord (approved):

PR #129, opened 2023-01-27
Feat(cli): add non-interactive use of metricize

eloy-encord:
Now, users can add metric names to `metricize` command with the optional input `--add` (`-a`) in case they already know the name of the metrics they want to execute. This implementation allows the user to run the `metricize` command in a non-interactive fashion keeping it simple for the user to rerun the same metrics.

### DIFF
--- a/src/encord_active/cli/main.py
+++ b/src/encord_active/cli/main.py
@@ -269,12 +269,13 @@ def metricize(
     target: Path = typer.Option(
         Path.cwd(), "--target", "-t", help="Directory of the project to run the metrics on.", file_okay=False
     ),
-    metric_names: Optional[list[str]] = typer.Option(
-        None,
+    metric_names: list[str] = typer.Option(
+        [],
         "--add",
         "-a",
         help="Name of the metric to run. This option can be repeated to add multiple metrics to the run.",
     ),
+    run_all: bool = typer.Option(False, "--all", help="Run all available metrics."),
     fuzzy: bool = typer.Option(
         False, help="Enable fuzzy search in the selection. (press [TAB] to select more than one) 🪄"
     ),
@@ -292,9 +293,11 @@ def metricize(
     )
 
     metrics = list(map(load_metric, get_metrics()))
+    if run_all:  # User chooses to run all available metrics
+        selected_metrics = metrics
 
     # (interactive) User chooses some metrics via CLI prompt selection
-    if metric_names is None:
+    elif not metric_names:
         choices = list(map(lambda m: Choice(m, name=m.TITLE), metrics))
         Options = TypedDict("Options", {"message": str, "choices": List[Choice], "vi_mode": bool})
         options: Options = {
--- a/src/encord_active/cli/main.py
+++ b/src/encord_active/cli/main.py
@@ -266,14 +266,9 @@ def quickstart(
 @cli.command()
 @ensure_project
 def metricize(
+    metric_names: Optional[list[str]] = typer.Argument(None, help="Names of the metrics to run."),
     target: Path = typer.Option(
         Path.cwd(), "--target", "-t", help="Directory of the project to run the metrics on.", file_okay=False
-    ),
-    metric_names: list[str] = typer.Option(
-        [],
-        "--add",
-        "-a",
-        help="Name of the metric to run. This option can be repeated to add multiple metrics to the run.",
     ),
     run_all: bool = typer.Option(False, "--all", help="Run all available metrics."),
     fuzzy: bool = typer.Option(
--- a/src/encord_active/cli/main.py
+++ b/src/encord_active/cli/main.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import List, Set, TypedDict
+from typing import List, Optional, Set, TypedDict
 
 import click
 import rich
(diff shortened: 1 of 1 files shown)


eloy-encord:
what about adding a `--all` option that runs all the metrics? It seems like `metricize` command should be the right place to run all the metrics.
Do we have this feature already somewhere else?

src/encord_active/cli/main.py
```diff
@@ -269,8 +269,14 @@ def metricize(
     target: Path = typer.Option(
         Path.cwd(), "--target", "-t", help="Directory of the project to run the metrics on.", file_okay=False
     ),
+    metric_names: Optional[list[str]] = typer.Option(
```
Encord-davids, line 272: should we make this an argument instead of an option? (mostly since `add` doesn't sound right for this)

Encord-davids: also, would it make more sense to provide a python module rather than a name?  this way, when users create new metrics they can just point to them

Encord-davids:
> what about adding a `--all` option that runs all the metrics? It seems like `metricize` command should be the right place to run all the metrics. Do we have this feature already somewhere else?

Yep, an `--all` option totally makes sense

src/encord_active/cli/main.py
eloy-encord: I think that the python module probably is a different functionality that the one coded as these don't appear in the interactive selection, perhaps with another option like `from`.

eloy-encord: qq: if we move it to be an argument instead of an option that would enforce the required check, right? If that's the case then the interactive part would be cancelled?

Encord-davids: > qq: if we move it to be an argument instead of an option that would enforce the required check, right? If that's the case then the interactive part would be cancelled?

Yep



> I think that the python module probably is a different functionality that the one coded as these don't appear in the interactive selection, perhaps with another option like `from`.

I see what you mean, but it also doesn't make sense for the user to know the names 🤷‍♂️

eloy-encord: fixed an issue that wasn't allowing the interactive version to run, now one should be able to access both interactive and non-interactive `metricize`.

eloy-encord: when you mention python module do you refer to the python script where the metric resides?

eloy-encord: > > qq: if we move it to be an argument instead of an option that would enforce the required check, right? If that's the case then the interactive part would be cancelled?
> 
> We can make the argument optional 🙂
> 
> > I think that the python module probably is a different functionality that the one coded as these don't appear in the interactive selection, perhaps with another option like `from`.
> 
> I see what you mean, but it also doesn't make sense for the user to know the names 🤷‍♂️

Let's denote the two different use cases that we need to tackle:
1. The user has been running the same metric subset of the available EA metrics via `metricize` but it's taking too much effort to interactively select each one of these metrics from the CLI checkbox each time they are needed. At the end, this type of user needs some way to state the metrics it wants to run within a single command. Current example of this type of user: Eloy.
2. The user has coded its own metrics, highly likely outside the metrics folder in EA source code as default EA installation is in the form of a `pip install`, and now it wants to run those metrics in a EA project. The first and only solution that we have implemented and comes to my mind right now is to call `execute_metrics` from an user made script (this approach is displayed in the example metric). This is OK, but we also want to allow to run arbitrary metrics via the CLI which is not possible atm. This is where a python module path and/or `from` optional input can be used in `metricize`. Example of this type of user: those who install EA via pip and want to run their own metrics.

eloy-encord: Point (1) also covers all the users that are trying to make a (semi) automated pipeline without recurring to python scripting.

Encord-davids: We need to find a way to treat 1 and 2 in the same. Essentially, EA metrics and user-built metrics are similar in essence, they are both python modules. I'm not convinced that typing titles of metrics, which is something we can easily change, is a good idea (especially in automated pipelines).

@frederik-encord what are your two cents here?

eloy-encord: If we want to treat (1) and (2) as the same, a good course of action would be to actually have different sets of metrics for each project as metrics serve on a project basis. I would decline to add file paths as inputs on methods that are used often like `metricize` (you want to run a metric not to remember its path every now and then). So, metrics should be like first class citizen where you can choose to add them to projects and when you run `metricize` on a project then it displays the metrics available for such project. Idea here is that it should be easy for both automated and not automated pipelines. I would only use file paths here to set a metric in a project.

eloy-encord: Regarding the use of metric names, we could do the same as WANDB and use metric hashes which we don't currently support. These hashes would be definitively better than names or paths that could change at any moment.

Encord-davids: > you want to run a metric not to remember its path every now and then

hashes would make this even worse 😅 

> So, metrics should be like first-class citizen where you can choose to add them to projects and when you run metricize on a project then it displays the metrics available for such project

I 100% agree, what do you say about implementing this now? This could be done relatively simply.

I suggest we keep the interactive version as the default when it's the first time the command is run or if a flag is provided (maybe `--save-choices`). We dump the choices into a file and use them going on.

To still allow running only specific metrics we also expose a `--interactive` flag.

eloy-encord: We should not dump that much functionality to `metricize` and leave as it is, we should add the project metrics elsewhere. On another point, I think that we might be deviating from the original intention of the PR that is to ease run and testing of metrics which already does compared to previous iteration.

Encord-davids: Makes sense, anyway happy to merge this and continue iterating over this in the future

Encord-davids (approved):

PR #131, opened 2023-01-30
feat(labels): adds label importing flow for simple classifications

Encord-davids:


src/encord_active/app/label_onboarding/label_onboardin.py
```diff
@@ -0,0 +1,99 @@
...
+        paths = [Path(data_unit["data_link"]) for data_units in lr_data_units for data_unit in data_units.values()]
+        image_to_class_map = {path.name: path.parent.stem for path in paths}
+        classes = set(image_to_class_map.values())
+        class_names_string = ", ".join(f'"{name}"' for name in classes)
+
+        ontology = create_ontology_structure(classes)
+
+        st.subheader(f"We have identified {len(classes)} classes: {class_names_string}")
+    else:
+        st.warning(f"We currently don't support {label_type}")
```
frederik-encord, line 55:
```suggestion
        st.warning(f"We currently don't support easy import of {label_type}")
```

src/encord_active/app/label_onboarding/label_onboardin.py
```diff
@@ -0,0 +1,99 @@
...
+        return
+
+    with st.form("label_import_form"):
+        st.write("Press submit if you would like to continue with these classes and create the following things:")
+        st.expander("Ontology structure").json(ontology.to_dict())
+        with st.expander("Metrics that will be executed"):
```
frederik-encord, line 61:
```suggestion
        with st.expander("What metrics do you want to run on your labels?"):
```

src/encord_active/app/label_onboarding/label_onboardin.py
```diff
@@ -0,0 +1,99 @@
...
+
+        ontology = create_ontology_structure(classes)
+
+        st.subheader(f"We have identified {len(classes)} classes: {class_names_string}")
+    else:
+        st.warning(f"We currently don't support {label_type}")
+        return
+
+    with st.form("label_import_form"):
+        st.write("Press submit if you would like to continue with these classes and create the following things:")
```
frederik-encord, line 59:
```suggestion
        st.write('Press "Import Labels" continue with these classes and create the following things:')
```

src/encord_active/app/label_onboarding/label_onboardin.py
```diff
@@ -0,0 +1,99 @@
...
+        st.expander("Ontology structure").json(ontology.to_dict())
+        with st.expander("Metrics that will be executed"):
+            selected_metrics = render_metric_selection(label_type)
+
+        if st.form_submit_button("Generate Labels"):
```
frederik-encord, line 64:
```suggestion
        if st.form_submit_button("Import Labels"):
```

src/encord_active/app/label_onboarding/label_onboardin.py
```diff
@@ -0,0 +1,99 @@
...
+    st.header("Importing labels flow")
+    st.info("This is a questionaire that will help us import your labels")
+    st.write("")
+
+    label_type = st.selectbox("1. What is the label type?", [v.value for v in LabelType])
+
+    if not label_type:
+        return
+    elif label_type == LabelType.CLASSIFICATION:
+        st.info("For classifications we only support datasets where the data is split accros class directories.")
```
frederik-encord, line 42:
```suggestion
        st.info(
            """
For classifications, we only support datasets where the data is split accros class directories.
For example, the classes `cat`, `dog`, and `horse` would be picked up from this file structure:
[three backticks]
├── cat/
│   ├── 0.png
│   └── ...
├── dog/
│   ├── 0.png
│   └── ...
└── horse/
    ├── 0.png
    └── ...
[three backticks]
        """
        )
```

src/encord_active/app/label_onboarding/label_onboardin.py
```diff
@@ -0,0 +1,99 @@
...
+
+        project = Project(get_state().project_paths.project_dir).load()
+        lr_data_units = [lr["data_units"] for lr in project.label_rows.values()]
+        paths = [Path(data_unit["data_link"]) for data_units in lr_data_units for data_unit in data_units.values()]
+        image_to_class_map = {path.name: path.parent.stem for path in paths}
+        classes = set(image_to_class_map.values())
+        class_names_string = ", ".join(f'"{name}"' for name in classes)
```
frederik-encord, line 49:
```suggestion
        class_names_string = ", ".join(f'`{name}`' for name in classes)
```

src/encord_active/lib/labels/classification.py
```diff
@@ -0,0 +1,98 @@
...
+        manualAnnotation=True,
+    )
+
+
+def update_label_row_with_classification(
+    label_row: LabelRow, classification: Classification, image_class: str
+) -> LabelRow:
+    attribute = classification.attributes[0]
+    classification_hash = str(uuid.uuid4())[:8]
+    data_hash = list(label_row["data_units"].keys())[0]
```
frederik-encord, line 69: Seems like this only works for single images? Should we perhaps at least raise an error of more data units are available?

frederik-encord (changes requested):
When I try to use this, I don't get the balloons 😞 
It seems to be because the embedding file already exists and is therefor skipped.
Screenshot 2023-01-31 at 10 28 31

I also think that the UI could use a helping hand from @mavis-encord 😅

Encord-davids:
> When I try to use this, I don't get the balloons 😞 It seems to be because the embedding file already exists and is therefor skipped. Screenshot 2023-01-31 at 10 28 31

Yeah, that makes sense since we don't re-create embeddings. I'll do two things here to mitigate that: 
1. only run data-related metrics in the init command
2. prevent saving empty embedding files  

> I also think that the UI could use a helping hand from @mavis-encord 😅 .. but we can do this in a later iteration.

@mavis-encord can make a lot of things better 🙇

src/encord_active/lib/labels/classification.py
Encord-davids: This is consistent with the way we import data units in the `init` command (we don't create image groups).
 
But you're right, an error would be helpful here in case we start importing more than one data unit

Encord-davids: Alternatively, we can add a `data_hash` optional argument and when that is empty we just add the class to all data units.

src/encord_active/lib/db/merged_metrics.py
```diff
@@ -56,7 +56,7 @@ def build_merged_metrics(metrics_path: Path) -> pd.DataFrame:
             main_df_objects.rename(columns={"score": f"{meta['title']}"}, inplace=True)
 
     main_df = pd.concat([main_df_images, main_df_objects, main_df_image_quality])
-    main_df["tags"] = None
+    main_df["tags"] = [[]] * len(main_df)
```
frederik-encord, line 59: This is dangerous! Doing this would use the reference to the same list for all data items. 
```suggestion
    main_df["tags"] = [[] for _ in range(len(main_df))]
```

frederik-encord (changes requested):

frederik-encord (approved):

PR #142, opened 2023-02-02
refactor(predictions): add types without breaking backward compatibility

frederik-encord:


Encord-davids (approved):
Thank you for this! Don't know how I could understand anything without these changes 💪

PR #145, opened 2023-02-02
Feat (metric): image singularity

Gorkem-Encord:
This metric gives each image a score that shows each image's uniqueness.
- A score of zero means that the image has duplicates in the dataset; on the other hand, a score close to one represents that image is quite unique. Among the duplicate images, we only give a non-zero score to a single image, and the rest will have a score of zero (for example, if there are five identical images, only four will have a score of zero). This way, these duplicate samples can be easily tagged and removed from the project.    
- Images that are near duplicates of each other will be shown side by side. 
### Possible actions
- **To delete duplicate images:** You can set the quality filter to cover only zero values (that ends up with all the duplicate images), then use bulk tagging (e.g., with a tag like `Duplicate`) to tag all images.
- **To mark duplicate images:** Near duplicate images are shown side by side. Navigate through these images and mark whichever is of interest to you.

### DIFF
--- a/src/encord_active/lib/metrics/metric.py
+++ b/src/encord_active/lib/metrics/metric.py
@@ -122,7 +122,7 @@ class Metric(ABC):
         long_description: str,
         metric_type: MetricType,
         data_type: DataType,
-        annotation_type: List[Union[ObjectShape, ClassificationType]],
+        annotation_type: Optional[List[Union[ObjectShape, ClassificationType]]],
         embedding_type: Optional[EmbeddingType] = None,
     ):
         self.metadata = MetricMetadata(
--- a/src/encord_active/lib/metrics/metric.py
+++ b/src/encord_active/lib/metrics/metric.py
@@ -122,7 +122,7 @@ class Metric(ABC):
         long_description: str,
         metric_type: MetricType,
         data_type: DataType,
-        annotation_type: Optional[List[Union[ObjectShape, ClassificationType]]],
+        annotation_type: List[Union[ObjectShape, ClassificationType]] = [],
         embedding_type: Optional[EmbeddingType] = None,
     ):
         self.metadata = MetricMetadata(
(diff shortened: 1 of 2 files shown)


Encord-davids (dismissed):

src/encord_active/lib/metrics/metric.py
```diff
@@ -122,7 +122,7 @@ def __init__(
         long_description: str,
         metric_type: MetricType,
         data_type: DataType,
-        annotation_type: List[Union[ObjectShape, ClassificationType]],
+        annotation_type: Optional[List[Union[ObjectShape, ClassificationType]]],
```
Encord-davids, line 125: Since this is a list, we should probably just default to an empty one instead of also dealing with `None`s everywhere.

```suggestion
        annotation_type: List[Union[ObjectShape, ClassificationType]] = [],
```

src/encord_active/lib/metrics/semantic/image_singularity.py
```diff
@@ -0,0 +1,115 @@
...
+- A score of zero means that the image has duplicates in the dataset; on the other hand, a score close to one represents that image is quite unique. Among the duplicate images, we only give a non-zero score to a single image, and the rest will have a score of zero (for example, if there are five identical images, only four will have a score of zero). This way, these duplicate samples can be easily tagged and removed from the project.    
+- Images that are near duplicates of each other will be shown side by side. 
+### Possible actions
+- **To delete duplicate images:** You can set the quality filter to cover only zero values (that ends up with all the duplicate images), then use bulk tagging (e.g., with a tag like `Duplicate`) to tag all images.
+- **To mark duplicate images:** Near duplicate images are shown side by side. Navigate through these images and mark whichever is of interest to you.
+""",
+            metric_type=MetricType.SEMANTIC,
+            data_type=DataType.IMAGE,
+            embedding_type=EmbeddingType.CLASSIFICATION,
+            annotation_type=None,
```
Encord-davids, line 43: If the suggestion above is applied this can be removed 

```suggestion
```
or explicitly set

```suggestion
            annotation_type=AnnotationType.NONE,
```

Encord-davids (dismissed):

Encord-davids (approved):

PR #146, opened 2023-02-02
chore(metrics-metadata): Move metrics metadata to separate dataclass

yogesh-encord:


src/encord_active/lib/metrics/metric.py
```diff
@@ -52,23 +53,54 @@ class AnnotationType:
...
-    annotation_type: Optional[List[AnnotationTypeUnion]]
-    embedding_type: Optional[str]
-    data_type: DataType
-    long_description: str
-    metric_type: MetricType
-    needs_images: bool
-    short_description: str
-    title: str
+@dataclass
+class StatsMetadata:
```
yogesh-encord, line 57: Only saw one usage of the min/max stats, so this is currently unused. Would like to know if the min/max/mean/threshold stats as part of metrics metadata are used anywhere.

src/encord_active/lib/metrics/example.py
```diff
@@ -28,6 +28,24 @@ class ExampleMetric(Metric):
 $$h_{\lambda}(x) = \frac{1}{x^\intercal x}$$
 """
 
+    def __init__(self):
```
eloy-encord, line 31: seems like there is a duplicated code section just above this __init__ function regarding title, metrci_type, ... Do you intend to remove that and move everything inside __init__ or else?

src/encord_active/lib/metrics/utils.py
```diff
@@ -55,12 +55,8 @@ def load_metric_dataframe(
     df = pd.read_csv(metric.path).sort_values([sorting_key, "identifier"], ascending=True).reset_index()
 
     if normalize:
-        min_val = metric.meta.get("min_value")
```
eloy-encord, line 58: My guess is that the intent with storing `min_value` and `max_value` in the metadata is to avoid recalculating this value each time you explore a metric in the UI. The difference should be noticed when the dataset is large enough.

src/encord_active/lib/metrics/example.py
yogesh-encord: Yep! done

src/encord_active/lib/metrics/utils.py
yogesh-encord: That makes sense, I'll add those to the StatsMetadata thing and make them available

src/encord_active/lib/metrics/metric.py
```diff
@@ -52,23 +53,54 @@ class AnnotationType:
...
+    ):
+        self.metadata = MetricMetadata(
+            title=title,
+            short_description=short_description,
+            long_description=long_description,
+            metric_type=metric_type,
+            data_type=data_type,
+            annotation_type=annotation_type,
+            embedding_type=embedding_type,
+            needs_images=False if not needs_images and metric_type == MetricType.GEOMETRIC else True,
```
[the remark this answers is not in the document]
yogesh-encord: Getting rid of it

src/encord_active/lib/db/merged_metrics.py
```diff
@@ -21,30 +23,20 @@ def build_merged_metrics(metrics_path: Path) -> pd.DataFrame:
...
             continue
 
-        if len(metric_scores["identifier"][0].split("_")) == 3:
-            # Image-level index
-            if main_df_images.shape[0] == 0:
-                columns_to_merge = metric_scores[["identifier", "url", "score"]]
-            else:
-                columns_to_merge = metric_scores[["identifier", "score"]]
-            main_df_images = pd.merge(main_df_images, columns_to_merge, how="outer", on="identifier")
-            main_df_images.rename(columns={"score": f"{meta['title']}"}, inplace=True)
```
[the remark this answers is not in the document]
yogesh-encord: This could be simplified, but not related to the current change so will remove

Encord-davids:
We should probably commit the lock file if we are making any changes to the specs

Encord-davids (approved):

PR #153, opened 2023-02-08
feat(metrics): Simple Metrics for faster execution

yogesh-encord:


src/encord_active/lib/metrics/heuristic/img_features.py
```diff
@@ -10,29 +10,25 @@
...
     Metric,
     MetricType,
+    SimpleMetric,
 )
 from encord_active.lib.metrics.writer import CSVMetricWriter
 
 
-def iterate_with_rank_fn(
-    iterator: Iterator,
+def execute_helper(
```
Encord-davids, line 18: [curious]🤔

Seems to be a lot of repetition using this method which makes all the executes the same essentially. 

Should this maybe be implemented by `SimpleMetric` and each subclass implements `execute` instead of e.g. `execute` and  `rank_by_contrast`?

Since all the `rank_by_****` are getting just an image, this should be the only argument execute needs no?

PR #154, opened 2023-02-09
feat(predictions): adds classification predictions

Encord-davids:


src/encord_active/lib/model_predictions/writer.py
```diff
@@ -44,16 +51,20 @@ class LabelEntry:
...
     class_id: ClassID  # TODO remove this stupid legacy thing.
-    x1: float = 0.0
-    y1: float = 0.0
-    x2: float = 0.0
-    y2: float = 0.0
+    x1: Optional[float] = None
+    y1: Optional[float] = None
+    x2: Optional[float] = None
+    y2: Optional[float] = None
     rle: Optional[RLEData] = None
```
yogesh-encord, line 58: Are these always present or absent together? Maybe having some sort of other structy thing to denote that could be nicer?

src/encord_active/app/common/components/prediction_grid.py
```diff
@@ -34,7 +34,7 @@ def build_card_for_labels(
     data_dir: Path,
     label_color: Color = Color.RED,
 ):
-    class_colors = {int(k): idx["color"] for k, idx in get_state().predictions.all_classes.items()}
+    class_colors = {int(k): idx.get("color", Color.RED) for k, idx in get_state().predictions.all_classes.items()}
```
yogesh-encord, line 37: is there another suitable name?

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -3,25 +3,32 @@
...
+)
 from encord_active.lib.project import Project
 
 logger = logging.getLogger(__name__)
 BBOX_KEYS = {"x", "y", "w", "h"}
 BASE_URL = "https://app.encord.com/label_editor/"
 
 
-class PredictionType(Enum):
+class PredictionType(str, Enum):
```
yogesh-encord, line 31: Any reason not to use StrEnum?

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -220,16 +270,48 @@ def __prepare_class_id_lookups(self, custom_map: Optional[Dict[str, int]]):
...
-        for obj in self.project.ontology.classifications:
-            self.classification_class_id_lookup[obj.feature_node_hash] = len(self.classification_class_id_lookup)
+        self.classification_class_id_lookup: Dict[FrameClassification, ClassID] = {}
+        for classification, attribute, option in iterate_classification_attribute_options(self.project.ontology):
+            key = FrameClassification(
+                classification_hash=classification.feature_node_hash,
+                attribute_hash=attribute.feature_node_hash,
+                option_hash=option.feature_node_hash,
+            )
+            self.classification_class_id_lookup[key] = len(self.classification_class_id_lookup)
```
yogesh-encord, line 280: could be nicer a dict comprehension using enumerate maybe?

src/encord_active/lib/project/project.py
```diff
@@ -62,6 +62,11 @@ def load(self, subset_size: Optional[int] = None) -> Project:
 
         return self
 
+    def iterate_data_units(self):
```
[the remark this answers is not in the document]
Encord-davids: Good catch 👍

src/encord_active/lib/model_predictions/writer.py
Encord-davids: Probably would, but that is a hole we should not dive into right now since these would have loads of implications. (we save these and load them to dataframes)

src/encord_active/app/common/components/prediction_grid.py
Encord-davids: what would you consider a suitable name?

src/encord_active/lib/model_predictions/writer.py
Encord-davids: In most cases multiline comprehensions are considered bad, in our case, it would look like

```python
        self.classification_class_id_lookup = {
            FrameClassification(
                classification_hash=classification.feature_node_hash,
                attribute_hash=attribute.feature_node_hash,
                option_hash=option.feature_node_hash,
            ): index
            for index, (classification, attribute, option) in enumerate(
                iterate_classification_attribute_options(self.project.ontology)
            )
        }
```

but I think we can extend the iterator and yield both the hashes object and the items object which will make the comprehension almost a single line and still nice and readable (and a few others).

```python
        self.classification_class_id_lookup = {
            key: index for index, (key, _) in enumerate(iterate_classification_attribute_options(self.project.ontology))
        }
```

yogesh-encord: That'd be great! I remember seeing similar comprehensions in other spots as well, might make all of them simpler

src/encord_active/app/common/components/prediction_grid.py
yogesh-encord: Something like ontology_object? idx especially in a loop is a bit confusing as it feels like it's an int

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -268,48 +344,68 @@ def append_object_label(du_hash: str, frame: int, o: dict, width: int, height: i
...
             for du_hash, du in lr["data_units"].items():
                 height = int(du["height"])
                 width = int(du["width"])
 
-                if "im" in data_type:  # img_group or image
+                if data_type in ["image_group", "image"]:
                     frame = int(du["data_sequence"])
-                    for obj in du["labels"]["objects"]:
-                        append_object_label(du_hash, frame, obj, width, height)
+                    for label in du["labels"]["objects"]:
```
frederik-encord, line 354:
```suggestion
                    for label in du["labels"].get("objects", []):
```

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -268,48 +344,68 @@ def append_object_label(du_hash: str, frame: int, o: dict, width: int, height: i
...
+                        append_object_label(du_hash, frame, label, width, height)
+                    for label in du["labels"]["classifications"]:
```
frederik-encord, line 356:
```suggestion
                    for label in du["labels"].get("classifications", []):
```

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -268,48 +344,68 @@ def append_object_label(du_hash: str, frame: int, o: dict, width: int, height: i
...
+                        append_classification_label(du_hash, frame, label, answers)
                 else:  # Video
                     for fr, labels in du["labels"].items():
                         frame = int(fr)
-                        for obj in labels["objects"]:
-                            append_object_label(du_hash, frame, obj, width, height)
+                        for label in labels["objects"]:
```
frederik-encord, line 361:
```suggestion
                        for label in labels.get("objects", []):
```

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -268,48 +344,68 @@ def append_object_label(du_hash: str, frame: int, o: dict, width: int, height: i
...
+                            append_object_label(du_hash, frame, label, width, height)
+                        for label in labels["classifications"]:
```
frederik-encord, line 363:
```suggestion
                        for label in labels.get("classifications", []):
```

frederik-encord (changes requested):
I haven't had the time to actually try this out in practice, but will do today.

src/encord_active/app/common/components/prediction_grid.py
Encord-davids: ohh you meant `idx`, I thought you talked about `class_color` 🤦‍♂️

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -211,11 +229,26 @@ def import_KITTI_labels(
...
-                "w": row["xmax"] - row["xmin"],
-                "h": row["ymax"] - row["ymin"],
-            }
-            conf = row["undefined0"]
-            prediction_writer.add_prediction(data_hash, class_name, confidence_score=conf, bbox=bbox)
+            bbox = BoundingBox(
+                x=float(row["xmin"]),
+                y=float(row["ymin"]),
+                w=row["xmax"] - row["xmin"],
+                h=row["ymax"] - row["ymin"],
```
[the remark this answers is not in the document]
Encord-davids: we can make all of them `float` for explicitness, but since the `-` operation it is cast to a number anyway.

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -211,11 +229,26 @@ def import_KITTI_labels(
...
+            )
+            prediction_writer.add_prediction(
+                Prediction(
+                    data_hash=data_hash,
+                    confidence=float(row["undefined0"]),
+                    object=ObjectDetection(
+                        format=Format.BOUNDING_BOX, data=BoundingBox.parse_obj(bbox), object_class_hash=class_name
```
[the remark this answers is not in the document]
Encord-davids: that is actually not needed anymore, it used to be a dictionary 👍

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -390,77 +477,86 @@ def add_prediction(
...
+                    x1, y1, w, h = cv2.boundingRect(polygon)  # type: ignore
+                else:  # Polygon is points
+                    # Read image size from label row
+                    if np.issubdtype(polygon.dtype, np.integer):
+                        polygon = polygon.astype(float) / np.array([[width, height]])
+
+                    np_mask = points_to_mask(polygon, width=width, height=height)  # type: ignore
+                    x1, y1, w, h = cv2.boundingRect((polygon * np.array([[width, height]])).reshape(-1, 1, 2).astype(int))  # type: ignore
+                x2, y2 = x1 + w, y1 + h
+                rle = binary_mask_to_rle(np_mask)
```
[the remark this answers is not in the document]
Encord-davids: I initially wanted to change this and would definitely make it easier to read. However, each branch relies on the current conditions for proper typing. Since I couldn't find a way to tell python the types are matching, we can change `Prediction` to not have `format` and `data` and instead have data of type `Polygon`, `Mask` or `BoundingBox` (but I think this isn't in the scope of this PR).

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -390,77 +477,86 @@ def add_prediction(
...
+
+            else:
+                class_id = self.object_class_id_lookup.get(prediction.object.object_class_hash)
```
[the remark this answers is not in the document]
Encord-davids: yes 👍

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -390,77 +477,86 @@ def add_prediction(
...
+                bbox = prediction.object.data
+                ptype = PredictionType.BBOX
+                # TODO: move me to the prediction model
+                self.__check_bbox(bbox)
```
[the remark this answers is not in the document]
Encord-davids: totally right 👍

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -390,77 +477,86 @@ def add_prediction(
...
+                x1 = bbox.x * width
+                y1 = bbox.y * height
+                x2 = (bbox.x + bbox.w) * width
+                y2 = (bbox.y + bbox.h) * height
+
+            ontology_object = self.object_lookup[prediction.object.object_class_hash]
```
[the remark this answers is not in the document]
Encord-davids: I don't think this is right, `object_lookup` keys are the [hashes](https://github.com/encord-team/encord-active/blob/9252d419f6402d029798e124e8854acbba8c7bd5/src/encord_active/lib/model_predictions/writer.py#L230).

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -390,77 +477,86 @@ def add_prediction(
...
             data_unit = label_row["data_units"][data_hash]
             if "data_sequence" in data_unit:
                 _frame = int(data_unit["data_sequence"])
 
         object_hash = self.__get_unique_object_hash()
 
-        class_id = self.object_class_id_lookup.get(class_uid)
         if class_id is None:
             raise ValueError(
                 f"`class_uid` didn't match any key in the "
```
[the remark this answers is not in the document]
Encord-davids: Don't think I understand what you mean 😅

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -390,77 +477,86 @@ def add_prediction(
...
                 y2=y2,
                 rle=rle,
             )
         )
+
+    def add_object_prediction(self):
+        pass
+
+    def add_classification_prediction(self):
+        pass
```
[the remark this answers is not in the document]
Encord-davids: funny, I don't recall adding these 😅

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -268,48 +344,68 @@ def append_object_label(du_hash: str, frame: int, o: dict, width: int, height: i
 
         for lr in tqdm(self.project.label_rows.values(), desc="Preparing labels", leave=True):
             data_type = lr["data_type"]
+            answers = lr["classification_answers"]
             for du_hash, du in lr["data_units"].items():
                 height = int(du["height"])
                 width = int(du["width"])
 
-                if "im" in data_type:  # img_group or image
+                if data_type in ["image_group", "image"]:
```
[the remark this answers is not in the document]
Encord-davids: ohh good catch 👍

src/encord_active/lib/model_predictions/writer.py
Encord-davids: I think the keys are always there, but good point, will change for safety 👍

src/encord_active/lib/model_predictions/writer.py
Encord-davids: I think the keys are always there, but good point, will change for safety 👍

docs/docs/sdk/importing-model-predictions.mdx
```diff
@@ -80,18 +112,20 @@ That is:
...
     frame = 3,  # optional frame for videos
-    class_id = "<the_class_id>",
     confidence = 0.8,
-    format = Format.BOUNDING_BOX,
-    # highlight-start
-    # Your bounding box coordinates in relative terms (% of image width/height).
-    data = BoundingBox(x=0.2, y=0.3, w=0.1, h=0.4)
-    # highlight-end
+    object=ObjectDetection(
+        object_class_hash="<the_class_id>",
```
frederik-encord, line 122: `object_class_hash` as well as `classification_hash` could be very confusing as the entry in the ontology is `featureNodeHash` and in the objects `featureHash` so perhaps just call it feature_hash or something completely new?

frederik-encord (changes requested):
One thing that I noticed is that the filter function in `run_all_prediction_metrics` tries to access m.metadata on the class and not the instance in `src/encord_active/lib/metrics/execute.py`. This is a regression from the metric refactor but still breaks this PR.
We need to do `m().metadata` instead.

src/encord_active/lib/model_predictions/writer.py
```diff
@@ -351,35 +459,14 @@ def __check_bbox(bbox):
             raise ValueError("Bbox coordinates should be floats")
         return True
```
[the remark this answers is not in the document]
Encord-davids: There was one more reference to `__check_bbox` which I now also removed 👍

docs/docs/sdk/importing-model-predictions.mdx
Encord-davids: Good point, I'll rename that

frederik-encord (dismissed):

frederik-encord (approved):

frederik-encord (approved):

PR #156, opened 2023-02-09
[FIX] replace is used instead of rename

feemthan:
This fix is because recent pandas update throws warnings while running the app, pandas requests replace to be used instead of rename in their documentation.

### DIFF
--- a/src/encord_active/app/common/components/metric_summary.py
+++ b/src/encord_active/app/common/components/metric_summary.py
@@ -64,12 +64,12 @@ def render_2d_metric_plots(metrics_data_summary: MetricsSeverity):
         x_metric_df = metrics_data_summary.metrics[str(x_metric_name)].df[
             [MetricWithDistanceSchema.identifier, MetricWithDistanceSchema.score]
         ]
-        x_metric_df.rename(columns={MetricWithDistanceSchema.score: f"{CrossMetricSchema.x}"}, inplace=True)
+        x_metric_df.columns = x_metric_df.columns.str.replace(f'{MetricWithDistanceSchema.score}', f'{CrossMetricSchema.x}')
 
         y_metric_df = metrics_data_summary.metrics[str(y_metric_name)].df[
             [MetricWithDistanceSchema.identifier, MetricWithDistanceSchema.score]
         ]
-        y_metric_df.rename(columns={MetricWithDistanceSchema.score: f"{CrossMetricSchema.y}"}, inplace=True)
+        y_metric_df.columns = y_metric_df.columns.str.replace(f'{MetricWithDistanceSchema.score}', f'{CrossMetricSchema.y}')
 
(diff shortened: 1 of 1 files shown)


Encord-davids:
> pandas requests replace to be used instead of rename in their documentation

Could you add a reference and specify the offending `pandas` version? 🙏 
So far we have successfully been using `rename` and if that is not supported anymore we would need to update all usages of it in addition to the one you are suggesting.

One caveat with `replace` is that it will also apply to substrings which might open the door for bugs 😕
E.g. we have two columns `score` and `new_score` by calling `str.column.replace("score", "old_score")` we would end up with the following columns: `["old_score", "old_score_new"]` which isn't what we want.

feemthan:
Aah, I see the problem. The culprit is inplace = True then. This seems to work for me and will update all instances of rename.

Encord-davids:
> Aah, I see the problem. The culprit is inplace = True then. This seems to work for me and will update all instances of rename.

I think there might be an even deeper culprit which is the fact that we create a [view instead of a copy](https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy). So the warning is raised when you rename (or similar operations) in place on a view.

Since renaming isn't the only operation we do on `x_metric_df` and `y_metric_df` we will need to make a copy of them anyway ([and we do](https://github.com/encord-team/encord-active/blob/d5da934599103f32d60477ed05e17d0436012aff/src/encord_active/app/common/components/metric_summary.py#L87-L91)).

I suggest we just make a copy in the first place, which should resolve all issues. 
```python
x_metric_df = (
    metrics_data_summary.metrics[str(x_metric_name)]
    .df[[MetricWithDistanceSchema.identifier, MetricWithDistanceSchema.score]]
    .copy()
)
```

feemthan:
@Encord-davids I did think about this, but wanted to make sure that this does not memory overload the server as copying is always a red flag.

Encord-davids:
> @Encord-davids I did think about this, but wanted to make sure that this does not memory overload the server as copying is always a red flag.

Good point, but since we already use a copy at a later stage I think this is fine (we'll deal with memory issues if they arise in the future)

feemthan:
Ok please feel free to close the pull request then.

Encord-davids:
> Ok please feel free to close the pull request then.

@feemthan thanks a lot for raising this and for flagging this and going through the effort of suggesting a fix 🙌 
Really looking forward to more feedback and contributions 🙏 

Closing this in favour of [PR-160](https://github.com/encord-team/encord-active/pull/160)

Encord-davids:
> Ok please feel free to close the pull request then.

@feemthan thanks a lot for raising this and for flagging this and going through the effort of suggesting a fix 🙌
Really looking forward to more feedback and contributions 🙏

Closing this in favour of https://github.com/encord-team/encord-active/pull/160

PR #165, opened 2023-02-13
chore: disable file watchers when not in development mode

frederik-encord:
Disable Streamlit file watchers unless `dev` is set to true in the EA config.
`encord-active config set dev true`


src/encord_active/app/app_config.py
```diff
@@ -36,6 +37,10 @@ def load(self):
 
         self.contents = toml.load(self.config_file)
 
+    @property
+    def is_dev(self) -> bool:
+        return self.contents.get(ConfigProperties.DEV.value, False)
```
yogesh-encord, line 42: Should we do the opposite and set DEV to true by default? Doesn't really really matter here but maybe we'll start using this for more things.

frederik-encord: I don't want production users to get the `inotify too many watchers` error when running the app, which is why I set the default to False to avoid that people get that error.

frederik-encord: @yogesh-encord what do you think?

yogesh-encord: Makes sense, we can look at it again if we start using it for other things

yogesh-encord (approved):

yogesh-encord (approved):

PR #167, opened 2023-02-13
fix: handle empty data_unit labels

Gorkem-Encord:


src/encord_active/lib/model_predictions/writer.py
```diff
@@ -274,7 +274,7 @@ def append_object_label(du_hash: str, frame: int, o: dict, width: int, height: i
 
                 if "im" in data_type:  # img_group or image
                     frame = int(du["data_sequence"])
-                    for obj in du["labels"]["objects"]:
+                    for obj in du["labels"].get("objects", []):
```
frederik-encord, line 277: Same thing needs to happen in line 282.

frederik-encord (changes requested):

src/encord_active/lib/model_predictions/writer.py
frederik-encord: And preferably make the same change if there is any other place in the code where we assume that the "objects" list exists.

Gorkem-Encord: 👍 did some related searches and fixed the same situation in also other places.

frederik-encord (approved):

PR #168, opened 2023-02-14
fix(metrics-metadata): Check for index_type in addition to metrics_type

yogesh-encord:


src/encord_active/lib/metrics/utils.py
```diff
@@ -124,7 +124,7 @@ def load_metric_metadata(meta_pth) -> MetricMetadata:
             short_description=old_meta["short_description"],
             long_description=old_meta["long_description"],
             data_type=old_meta["data_type"],
-            metric_type=old_meta["metric_type"],
+            metric_type=old_meta["metric_type"] if "metric_type" in old_meta else old_meta["index_type"],
```
Encord-davids, line 127: In case `index_type` is also not there, maybe something like:

```suggestion
            metric_type=old_meta.get("metric_type", old_meta.get("index_type"))

Encord-davids (dismissed):

Encord-davids (approved):

PR #171, opened 2023-02-14
misc: show exact metric name when logging it

eloy-encord:


src/encord_active/lib/metrics/execute.py
```diff
@@ -140,7 +140,10 @@ def _execute_metrics(cache_dir: Path, iterator: Iterator, metrics: list[Metric])
 
 
 def _execute_simple_metrics(cache_dir: Path, iterator: Iterator, metrics: list[SimpleMetric]):
-    logger.info(f"Running Metrics <blue>{', '.join(metric.metadata.title.title() for metric in metrics)}</blue>")
+    if len(metrics) == 0:
+        logger.info("No metric was chosen to run.")
```
yogesh-encord, line 144: I thin we should just not log anything here right?  Because it's possible that other metrics were run through the _execute_metrics

eloy-encord: As you are original code owner, should I remove it without worries?

src/encord_active/lib/metrics/execute.py
```diff
@@ -140,7 +140,10 @@ def _execute_metrics(cache_dir: Path, iterator: Iterator, metrics: list[Metric])
...
+        return
```
yogesh-encord, line 145:
```suggestion
    if len(metrics) == 0:
        return
```
I'd recommend changing to something like this!

src/encord_active/lib/metrics/execute.py
yogesh-encord: Yep

yogesh-encord (approved):

PR #177, opened 2023-02-15
fix(ui): make filters page not crash on missing network

yogesh-encord:


src/encord_active/app/actions_page/export_filter.py
```diff
@@ -156,7 +155,8 @@ def export_filter():
     message_placeholder = st.empty()
 
     st.header("Filter & Export")
-    action_utils, has_original_project = _get_project()
+    action_utils = _get_action_utils()
+    project_has_remote = action_utils.project_meta.get("has_remote", False)
```
Encord-davids, line 159: `action_utils` can be `None`, probably need some check here

Encord-davids (approved):

Encord-davids (approved):

PR #183, opened 2023-02-17
fix: more accurate point matching for 2D embedding plots

frederik-encord:


src/encord_active/app/data_quality/sub_pages/explorer.py
```diff
@@ -131,17 +131,26 @@ def build(self, selected_df: DataFrame[MetricSchema], metric_scope: MetricScope)
...
-    selected_points_merged = [str(int(item["x"] * 1000)) + "-" + str(int(item["y"] * 1000)) for item in selected_points]
-
-    selected_rows = embeddings_2d[embeddings_2d[Embedding2DSchema.x_y].isin(selected_points_merged)]
+    selection_raw = DataFrame[PointSelectionSchema](pd.DataFrame(selected_points))
+    matched = embeddings_2d.copy().merge(
+        selection_raw[[PointSelectionSchema.x, PointSelectionSchema.y]],
+        on=[Embedding2DSchema.x, Embedding2DSchema.y],
+        how="left",
+        indicator=True,
+    )
```
Gorkem-Encord, line 151: How about 'inner' merging here, which directly gets their intersection?

src/encord_active/app/data_quality/sub_pages/explorer.py
```diff
@@ -131,17 +131,26 @@ def build(self, selected_df: DataFrame[MetricSchema], metric_scope: MetricScope)
...
-    Here, we just create and additional column by merging x and y points, so that we can get the selected points'
-    identifiers for filtering.
-    """
-    selected_points_merged = [str(int(item["x"] * 1000)) + "-" + str(int(item["y"] * 1000)) for item in selected_points]
-
-    selected_rows = embeddings_2d[embeddings_2d[Embedding2DSchema.x_y].isin(selected_points_merged)]
+    selection_raw = DataFrame[PointSelectionSchema](pd.DataFrame(selected_points))
+    matched = embeddings_2d.copy().merge(
+        selection_raw[[PointSelectionSchema.x, PointSelectionSchema.y]],
+        on=[Embedding2DSchema.x, Embedding2DSchema.y],
```
Gorkem-Encord, line 148: Since we use different schemas here for the x and y points, what about creating a new point schema as a parent class for both of them here to avoid future bugs:

```
class PointSchema(pa.SchemaModel):
    x: Series[float] = pa.Field(coerce=True)
    y: Series[float] = pa.Field(coerce=True)

class PointSelectionSchema(PointSchema):
    curveNumber: Series[float] = pa.Field(coerce=True)
    pointNumber: Series[float] = pa.Field(coerce=True)
    pointIndex: Series[float] = pa.Field(coerce=True)
```

src/encord_active/lib/embeddings/utils.py
```diff
@@ -87,11 +87,21 @@ def _add_similarities(self, identifier: str):
...
+
+class PointSelectionSchema(PointSchema2D):
+    curveNumber: Series[float] = pa.Field(coerce=True)
+    pointNumber: Series[float] = pa.Field(coerce=True)
+    pointIndex: Series[float] = pa.Field(coerce=True)
+
+
+class Embedding2DSchema(IdentifierSchema, PointSchema2D):
     x: Series[float] = pa.Field(coerce=True)
     y: Series[float] = pa.Field(coerce=True)
```
Gorkem-Encord, line 103: these should be removed I think

Gorkem-Encord (approved):

PR #192, opened 2023-02-17
docs (metrics): update example metric

eloy-encord:


Gorkem-Encord (changes requested):
In these example projects, both image-level and object-level writing is happening in the same iteration loop. We do not allow such thing and it may confuse the users, so we should note somewhere that 'you can either create an image-level metric or an object-level metric, not both at the same time'. We can also divide the iterating over data_units into two and comment one of them to not cause any confusion.

frederik-encord:
What might want to do is having separate (sub)classes for image level and object level metrics and then enforce that they call the writer with the right parameters to avoid this confusion. Perhaps not in this PR though.

docs/docs/metrics/write-your-own.md
```diff
@@ -24,74 +24,85 @@ You can use the following template to get started with writing your own metric.
...
 class ExampleMetric(Metric):
     def __init__(self):
         super().__init__(
             title="Example Title",
             short_description="Assigns same value and description to all objects.",
-            long_description=r"""For long descriptions, I can use Markdown to _format_ the text.
+            long_description=r"""For long descriptions, you can use Markdown to _format_ the text.
 
-I can, e.g., make a
+E.g. you can make a
```
frederik-encord, line 43:
```suggestion
For example, you can make a
```

src/encord_active/lib/metrics/example.py
```diff
@@ -17,13 +17,13 @@ def __init__(self):
         super().__init__(
             title="Example Title",
             short_description="Assigns same value and description to all objects.",
-            long_description=r"""For long descriptions, I can use Markdown to _format_ the text.
+            long_description=r"""For long descriptions, you can use Markdown to _format_ the text.
 
-I can, e.g., make a
+E.g. you can make a
```
frederik-encord, line 22:
```suggestion
For example, you can make a
```

docs/docs/metrics/write-your-own.md
```diff
@@ -24,74 +24,85 @@ You can use the following template to get started with writing your own metric.
 Your implementation should call `writer.write(<object_score>, <object>)` for every object in the iterator or use `writer.write(<frame_score>)` for every data unit in the iterator.
```
frederik-encord, line 24:
```suggestion
Your implementation should call `writer.write(<object_score>, <object>)` for every object in the iterator **OR** use `writer.write(<frame_score>)` for every data unit in the iterator.
```

frederik-encord: Perhaps add a line saying (for now) that one should not do both in the same metric definition.

frederik-encord (changes requested):

frederik-encord (approved):

Gorkem-Encord (approved):

PR #199, opened 2023-02-21
feat: add metric management in projects

eloy-encord:


src/encord_active/lib/common/utils.py
```diff
@@ -54,6 +54,7 @@ class ProjectMeta(TypedDict):
     project_title: str
     ssh_key_path: str
     has_remote: bool
+    metrics: dict[str, str]
```
Encord-davids, line 57: can this keep the actual metric instead of the module name?

src/encord_active/cli/main.py
```diff
@@ -61,6 +63,7 @@ def list_commands(self, ctx: click.Context):
 cli.add_typer(config_cli, name="config", help="[green bold]Configure[/green bold] global settings 🔧")
 cli.add_typer(import_cli, name="import", help="[green bold]Import[/green bold] Projects or Predictions ⬇️")
 cli.add_typer(print_cli, name="print")
+cli.add_typer(metric_cli, name="metric", help="[green bold]Manage[/green bold] project's metrics.")
```
Encord-davids, line 66: I think it would make sense to move the `metricize` command into `metrics` and call it `run`

src/encord_active/cli/metric.py
```diff
@@ -0,0 +1,148 @@
...
+    Add metrics to the project.
+
+    Add metrics from:
+    - local source python modules
+
+    If no metric name is provided then all metrics found in the python module will be added to the project.
+    """
+    rich.print("Inspecting provided module ...")
+    metric_names = [] if metric_name is None else metric_name
+    metric_names_set = set(metric_names)
```
Encord-davids, line 37: [nitpick] 💅 

```suggestion
    metric_names_set = set(metric_names or [])```

File: src/encord_active/cli/metric.py [hunk 4]
```diff
@@ -0,0 +1,148 @@
...
+        return
+    found_metrics_set = set(found_metrics)
+
+    # add all metrics from the python module if the input doesn't contain any metric name.
+    if len(metric_names) == 0:
+        metric_names = found_metrics.copy()
+        metric_names_set = found_metrics_set.copy()
+
+    has_conflicts = False
+    for m_name in metric_names:
```
Encord-davids, line 57: Why not iterate the `set` instead of the `list`?

Encord-davids (approved):

src/encord_active/cli/main.py
eloy-encord: perfect, I'm fond of doing that also 🍀.

src/encord_active/cli/metric.py
eloy-encord: To preserve ordering

src/encord_active/lib/common/utils.py
eloy-encord: I don't understand well this question.

src/encord_active/cli/metric.py
Encord-davids: so maybe a dictionary? (the keys are always in order)

src/encord_active/lib/common/utils.py
Encord-davids: I assume that the type here is `dict[str, str]` which maps from metric name to module? e.g. `"Blur": "src/encord_active/lib/metrics/heuristic/img_features.py"`. 

So what I'm suggesting is to change it to `dict[str, Metric]` or even `Set[Metric]`

If I'm wrong please correct me and disregard.

eloy-encord: From my pov, we can't store metrics instances per se because they are not persistent. But we can try and add module copies inside a folder in the project and open those instead of the "remote" ones (like what pip does). Also, next step would be upgrade (update) a metric and after that add metric versioning. I would like to allow at the first/second instance editable metrics so you don't have to update them every time you make a change (like pip install -e path/to/package). Main point is that we don't want to lose track from where the metric comes from, either to update it or to have it as editable (making it easier to download metrics from url afterwards).

On another note, highly likely metric metadata will migrate to be created at add time instead of run.

eloy-encord: atm, if we use of `Metric`/`SimpleMetric` in `fetch_project_meta` level (where any transformation from the I/O version to the actual instance should happen) then a very tangled cyclic dependency is created `I/O` -> `Metric` -> `Iterator` -> `Project` -> `fetch_project_meta` -> `I/O` & `Metric`. I tried to remove it in vain last Friday.
Btw, from my pov that cyclic dependency is the reason why the code of project's metadata is already scattered almost everywhere.

src/encord_active/cli/metric.py
eloy-encord: done, a refactoring made the usage of list + set version not needed as no deletion was required so a dict was indeed perfect for this use case.

src/encord_active/lib/common/utils.py
Encord-davids: > From my pov, we can't store metrics instances per se because they are not persistent. But we can try and add module copies inside a folder in the project and open those instead of the "remote" ones (like what pip does). Also, next step would be upgrade (update) a metric and after that add metric versioning. I would like to allow at the first/second instance editable metrics so you don't have to update them every time you make a change (like pip install -e path/to/package). Main point is that we don't want to lose track from where the metric comes from, either to update it or to have it as editable (making it easier to download metrics from url afterwards).

I'm not sure I'm on board with this idea. I don't think we should be implementing metric version management into the the project, instead, I would suggest providing the metrics as an actual dependency and publishing them as a separate package (which we anyways want to do for the sake of modularity)

Encord-davids: > atm, if we use of `Metric`/`SimpleMetric` in `fetch_project_meta` level (where any transformation from the I/O version to the actual instance should happen) then a very tangled cyclic dependency is created `I/O` -> `Metric` -> `Iterator` -> `Project` -> `fetch_project_meta` -> `I/O` & `Metric`. I tried to remove it in vain last Friday. Btw, from my pov that cyclic dependency is the reason why the code of project's metadata is already scattered almost everywhere.

I understand that problem, we should solve it, but not in the scope of this PR probably.

eloy-encord:
There is a slight change that I’m going to add in this PR and it’s to move metrics data from `project_meta.yaml` to its own yaml file (like `metrics_meta.yaml`). The main reason is to get faster module lookups if we store the class name of metrics and we also need metrics’ metadata for the CLI command `show` (even before running the metric for its first time that atm generates the metadata). I’m not going to change the metadata generation in the metrics folder in this PR, just forward that info to the corresponding metric yaml when a metric is attached to the project for premature and easier lookup of metrics and their metadata.

This change will serve to avoid cluttering `project_meta.yaml` with too much info not exactly related to it. I see this change as a similar version of `poetry.lock` that doesn't state the overall information in a `toml` file. I could be wrong here but probably moving this data to a different location will allow to grow from there while keeping distinct type of information as modular as possible.

eloy-encord:
Quotes on the above:
We can improve performance of metric extraction once we add a metric because we can get and store the class name of the metric with such metric title and, with it, do direct lookups to the module instead of iterate through all the classes looking for the same title once again.
On the premature side, atm the only place we are registering metrics metadata after a metric is ran because of the observers' job but that implies that we need to run the metric before being able to show its metadata (in the CLI `metric show` command) which is not great as we have access to the metrics’ metadata at add time (we just need to store it also).

src/encord_active/lib/common/utils.py
eloy-encord: I moved the metrics metadata from the project_meta file to its own file so we can discuss its structure there.

src/encord_active/cli/metric.py
```diff
@@ -0,0 +1,229 @@
...
+    has_conflicts = False
+    for title in metric_titles:
+        if title in found_metric_titles:
+            # input metric title was found in the python module
+            if title in metrics_meta:
+                old_module_path = metrics_meta[title]["location"]
+                if full_module_path == Path(old_module_path).expanduser().resolve():
+                    rich.print(f"Requirement already satisfied: {title} in '{module_path.as_posix()}'.")
+                else:
+                    rich.print(f"[red]ERROR: Conflict found in metric {title}. It points to '{old_module_path}'.[/red]")
```
frederik-encord, line 71:
```suggestion
                    rich.print(f"[red]ERROR: Conflict found in metric {title}. The same metric title already points to '{old_module_path}. Please change the metric title if you intend to have both metrics included.'.[/red]")
```

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+If you don't specify a metric, Encord Active will run all the metrics attached to the project.
+```shell
+encord-active metric run 
+```
+
+You can also run specific metrics by using their title:
+```shell
+encord-active metric run "Metric Title" "Another Metric Title"
+```
+
```
frederik-encord, line 60: Perhaps we should mention that you can do

```
encord-active metric run --fuzzy
```

to select metrics interactively?

src/encord_active/lib/metrics/io.py
```diff
@@ -0,0 +1,84 @@
+import importlib.util
+import inspect
+from pathlib import Path
+from typing import Callable, Optional, Union
+
+from encord_active.lib.metrics.execute import logger
+from encord_active.lib.metrics.metric import Metric, SimpleMetric
+
+
+def fill_metrics_meta_with_builtin_metrics(metrics_meta: dict):
```
frederik-encord, line 10: We always seem to be calling this with an empty dictionary. Perhaps at least make the parameter optional to avoid specifying that argument.

frederik-encord (changes requested):

frederik-encord:
The `metricize` command is also still existing.
Also, for existing projects, the predefined metrics won't be listed right? 🤷🏻

docs/docs/cli/metric-management.md
eloy-encord: you are right, I didn't mentioned it on purpose as I want to change the behaviour of this method (copy/pasted from `metricize`) in a posterior PR from interactive first, what happens now if we don't use any arguments, to interactive by request using a `--interactive`, `-i` flag and have `metric run` for running all the metrics. That's why I didn't write the docs about the current optional attributes.

eloy-encord:
For existing projects we have the choice to call `import project` again in the same folder and it will generate the required file with the predefined metrics or also run the existent `metricize` command that only runs predefined metrics.

I expect to remove `metricize` command going forward too, and/or discuss if we are going to maintain the command like we did with metricise for a few weeks while we give a warning of going to be deprecated.

eloy-encord:
btw, I also have in mind a better choice than calling `import project` in existing projects that's related to have something like an option `--group builtin` in `metric add` in order to be able to add all predefined metrics in one go.

src/encord_active/lib/metrics/io.py
eloy-encord: done 👍

src/encord_active/cli/metric.py
eloy-encord: added

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+## metric list
+
+The `metric list` command lists all the metrics attached to the project.
+
+Metrics are listed in a case-insensitive sorted order.
+```shell
+encord-active metric list
+```
+
+## metric show
```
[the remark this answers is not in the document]
eloy-encord: ignored while we set the right docs format for the CLI

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+
+Metrics constitute the foundation of Encord Active.
+This section documents all the available commands on how to manage and run metrics in your projects.
+
+:::info
+Make sure your shell's current working directory is that of an Encord Active project, or your command points to one with the `--target` global option.
+:::
+
+
+## metric add
```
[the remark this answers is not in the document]
eloy-encord: ignored while we set the right docs format for the CLI

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+
+The `metric add` command adds required metrics to the project and attaches them.
```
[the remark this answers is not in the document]
eloy-encord: I rephrased it to be clearer.
```
The `metric add` command adds metrics to the project.  
These metrics can be listed, removed and run on the project data via other CLI commands. 
```

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+The `metric list` command lists all the metrics attached to the project.
+
+Metrics are listed in a case-insensitive sorted order.
+```shell
+encord-active metric list
+```
+
+## metric show
+
+The `metric show` command shows information about one or more available metrics in the project.
```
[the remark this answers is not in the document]
eloy-encord: done 👍

docs/docs/cli/introduction.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+To get help from the CLI, simply call `encord-active` to see the complete list of commands,
+then `--help` combined with any of those can give you more information.
+
+## Global options
+
+* `--help`: Display help information.
+* `--target (-t)`: The working directory for the Encord Active command. Defaults to the current working directory.
+
+<DocCardList/>
```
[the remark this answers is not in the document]
eloy-encord: got it, fixed.

docs/docs/cli/_category_.json
```diff
@@ -1,9 +1,8 @@
 {
   "label": "Command Line Interface",
   "position": 5,
-  "collapsed": true,
   "link": {
-    "type": "generated-index",
-    "description": "How to use the command line interface (CLI)."
+    "type": "doc",
+    "id": "introduction"
```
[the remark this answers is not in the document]
eloy-encord: I found a problem that if I kept both `_category_.json` and `index.mdx` it would create a CLI page section inside the CLI section so you would be able to loop there. I fixed it via removing the `_category_.json` file and adding the required information in `index.mdx`. I followed the idea that's in the `index.mdx` at the docs start (the welcome file).

Let me know if that's ok for the automatic migration. Otherwise, I would need some help here to get it done.

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+encord-active metric run 
+```
+
+You can also run specific metrics by using their title:
+```shell
+encord-active metric run "Metric Title" "Another Metric Title"
+```
+
+
+## metric list
```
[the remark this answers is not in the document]
eloy-encord: ignored while we set the right docs format for the CL

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+
+If you try to add a metric that is already present, it will be skipped and you will receive a notification.  
+In case a metric title is not found in the python module, you will get an error.
+
+:::caution
+Some shells may treat square braces (`[` and `]`) as special characters. It is suggested to always quote arguments containing these characters to prevent unexpected shell expansion.
+:::
+
+
+## metric remove
```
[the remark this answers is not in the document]
eloy-encord: ignored while we set the right docs format for the CL

docs/docs/cli/metric-management.md
```diff
@@ -0,0 +1,78 @@
...
+:::
+
+
+## metric add
+
+The `metric add` command adds required metrics to the project and attaches them.
+
+You can add metrics from local python modules (`.py`).  
+If you don't specify a metric, Encord Active will add all the metrics in the python module to the project.
+```shell
```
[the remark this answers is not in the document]
eloy-encord: done

docs/docs/cli/metric-management.md
frederik-encord: Alright got it.

frederik-encord (dismissed):

eloy-encord:
Github needed a rebase to clear the cache and complete the checks.

Encord-davids (approved):

PR #201, opened 2023-02-21
Fix (coco import): images with exif orientation info

Gorkem-Encord:
Some images (especailly taken with mobile phone cameras) have orientation info in their metafile. We apply the orientation information and use the resulting image for Encord Active.


src/encord_active/cli/utils/coco.py
```diff
@@ -71,15 +72,19 @@ def import_coco_project(images_dir: Path, annotations_file: Path, target: Path,
         destination_dir=target,
         use_symlinks=use_symlinks,
     )
+    temp_folder = Path("temp_folder_for_saved_images")
```
rad-cord, line 75: You're probably better of using the [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) context manager

```
    with TemporaryDirectory() as tmp_dir:
        yield Path(tmp_dir).absolute()
```
this will auto-clean up even if an exception is thrown. Now you never have to remember to clean things up!

src/encord_active/lib/coco/importer.py
```diff
@@ -57,20 +55,18 @@ def upload_img(
         return None
     img_exif = img.getexif()
     if img_exif and (274 in img_exif):  # 274 corresponds to orientation key for EXIF metadata
-        temp_file_name = "temp_image" + file_path.suffix
+        temp_file_name = file_path.name
         img = ImageOps.exif_transpose(img)
-        img.save(temp_file_name)
+        img.save(temp_folder / temp_file_name)
```
frederik-encord, line 60: It's not clear why we need to pass around the temporary directory.
I would probably just use something like
```
import os
from tempfile import TemporaryDirectory
with TemporaryDirectory() as d:
    out_file = os.path.join(d, temp_file_name)
    img.save(out_file)
    try: ...
    except: ...
```
To avoid passing around the directories.

frederik-encord (changes requested):

src/encord_active/lib/coco/importer.py
Gorkem-Encord: The thing is that when we import a coco project and [create a dataset](https://github.com/encord-team/encord-active/blob/ab3eb0b7f6c22c2a29bf8f543e98f4d0d0482a74/src/encord_active/cli/utils/coco.py#L75) for that we just record the paths for each image, then we use those paths in the [create_project](https://github.com/encord-team/encord-active/blob/ab3eb0b7f6c22c2a29bf8f543e98f4d0d0482a74/src/encord_active/cli/utils/coco.py#L77) function, that's why we need to keep images until the end of project creation step. If we remove them just after [this](https://github.com/encord-team/encord-active/blob/ab3eb0b7f6c22c2a29bf8f543e98f4d0d0482a74/src/encord_active/lib/coco/importer.py#L65), they cannot be found later to create the project, if I am not mistaken.

src/encord_active/cli/utils/coco.py
Gorkem-Encord: 👍 Nice to learn this handy class.

frederik-encord (approved):

PR #202, opened 2023-02-22
fix: pydantic validators allow reuse

Encord-davids:


yogesh-encord (approved):
we should fix the underlying issue at some point

PR #211, opened 2023-02-24
fix(explorer): adds a divider between rows

Encord-davids:


frontend_components/src/encord_active_components/__init__.py
```diff
@@ -25,7 +25,7 @@ def pages_menu(items: List[MenuItem]) -> str:
 
 
 if __name__ == "__main__":
```
frederik-encord, line 27: Do we really want to keep this?

frederik-encord (approved):

PR #212, opened 2023-02-27
feat(project): allow choosing project from the ui

Encord-davids:


src/encord_active/app/streamlit_entrypoint.py
```diff
@@ -55,26 +59,41 @@ def to_items(d: dict, parent_key: Optional[str] = None):
...
+        projects = [target_path]
+    else:
+        parent_project = try_find_parent_project(target_path)
+        if parent_project:
+            projects = [parent_project]
+        else:
+            projects = find_child_projects(target_path)
+
+    with st.sidebar:
+        project_path = st.selectbox("Choose Project", projects, format_func=lambda path: path.name)
```
frederik-encord, line 77: Can we do the `selectbox` only when there are multiple projects?

src/encord_active/cli/utils/decorators.py
```diff
@@ -32,78 +32,94 @@ def find_child_projects(target: Path) -> List[Path]:
...
-            )
-        )
+            if not allow_multi:
+                choosen_target = choose_project(child_projects)
+                if not choosen_target:
+                    rich.print("No project was selected. Aborting.")
+                    raise typer.Exit()
 
-        raise typer.Exit()
+                kwargs["target"] = choosen_target
```
frederik-encord, line 119: I am not sure that this is something that we would want to do as a side-effect in this function? Shouldn't it be in the visualize definition instead?

frederik-encord (changes requested):

frederik-encord (approved):

PR #217, opened 2023-03-01
feat: add acquisition functions

eloy-encord:


eric-encord (commented):
One thing here is we are running a log function in the scoring function, but want to be sure that this is not going to create an error with zeros or negative numbers

eloy-encord:
You are right, I forgot to migrate np.nan_to_num() check from the playground script to the metric. With that function encapsulating the log the result will be the proper one as 0 * x = 0, where x is the substitute number in np.nan_to_num(). Negatives numbers shouldn't be expected as the input as a matrix of probabilities.

src/encord_active/lib/metrics/acquisition_functions.py
```diff
@@ -0,0 +1,169 @@
...
+            data_type=DataType.IMAGE,
+            annotation_type=AnnotationType.NONE,
+            model=model,
+        )
+
+    def score_predicted_class_probabilities(self, predictions: np.ndarray) -> float:
+        # silence divide by zero warning as the result will be correct (log2(0) is -inf, when multiplied by 0 gives 0)
+        # raise exception if invalid (negative) values are found in the predictions
+        with np.errstate(divide="ignore", invalid="raise"):
+            return -np.multiply(predictions, np.nan_to_num(np.log2(predictions))).sum(axis=1).mean()
```
Gorkem-Encord, line 100: It seems like this context manager only reports a generic error like `FloatingPointError`. It would be nice to report to the user explicitly that there are negative values in the probabilities.

src/encord_active/lib/metrics/acquisition_functions.py
```diff
@@ -0,0 +1,169 @@
...
+                continue
+            pred_probs = self.get_predicted_class_probabilities(img)
+            if pred_probs is None:
+                continue
+            score = self.score_predicted_class_probabilities(pred_probs)
+            writer.write(score)
+
+    def get_predicted_class_probabilities(self, image) -> Optional[np.ndarray]:
+        image_array = np.asarray(image).flatten()
+        return self._model.predict_proba([image_array])
```
Gorkem-Encord: I think checking the model probabilities should be performed in this function, just after feeding image to the model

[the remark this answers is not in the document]
eloy-encord: moving towards this today, wrap the expected methods for models behind a class

eloy-encord: I followed your advise and put it there (now it lives in the ModelWrapper class).

src/encord_active/lib/metrics/acquisition_functions.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Args:
+            data_path (Path): Path to the data sample.
+
+        Returns:
+            Data sample prepared to be used as input of `self.predict_probabilities()` method.
+        """
+        return [np.asarray(Image.open(data_path)).flatten()]
+
+    def predict_probabilities(self, data) -> Optional[np.ndarray]:
```
Gorkem-Encord, line 40: Do we need that function, can `predict_probabilities` handle that transform for simpler interface?

Gorkem-Encord:
What if we transform it like the following: `ModelWrapper` will be an abstract class with one/two methods including `predict_probabilities(data_path: Path)`. And it will be inherited by actual model classes like `class LogisticRegressionModel(ModelWrapper):`, `class ResNetModel(ModelWrapper):` etc. As an example, we can implement only the logistic regression thing, overall flow and the architecture is very good, I just want it to be more explicit for the user. This way we will be more explicitly saying that acquisition functions are agnostic to ML models, just implement the ModelWrapper class.

PR #220, opened 2023-03-02
feat(versioning): allows creation and navigation between versions

Encord-davids:


src/encord_active/lib/versioning/git.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @property
+    def versions(self):
+        return [_commit_to_version(commit) for commit in self.repo.iter_commits(self.repo.heads.main)]
+
+    @property
+    def has_changes(self):
+        return self.repo.is_dirty() or bool(self.repo.untracked_files)
+
+    def create_version(self, name: str):
```
frederik-encord, line 46: Since we expose this via the `lib`, should we raise an error if we're not on the latest version?

frederik-encord (approved):

src/encord_active/lib/versioning/git.py
Encord-davids: Yep, sounds like a good idea :+1:

frederik-encord (approved):

PR #225, opened 2023-03-03
fix: duplicate classifications in same label row crashes embeddings

frederik-encord:
fix: check that label row has classification answer

### DIFF
--- a/src/encord_active/lib/dataset/summary_utils.py
+++ b/src/encord_active/lib/dataset/summary_utils.py
@@ -1,4 +1,5 @@
 import json
+import logging
 from dataclasses import dataclass, field
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


src/encord_active/lib/dataset/summary_utils.py
```diff
@@ -93,6 +93,12 @@ def get_all_annotation_numbers(project_paths: ProjectFileStructure) -> Annotatio
 
                     for classification in data_unit["labels"].get("classifications", []):
                         classificationHash = classification["classificationHash"]
+                        if classificationHash not in label_row_meta["classification_answers"]:
+                            logger.warning(
```
Gorkem-Encord, line 97: Shall we import logger?

Encord-davids: Done

yogesh-encord (approved):

PR #231, opened 2023-03-07
Feat (metrics): classification performance metrics

Gorkem-Encord:
- Object and classification performance metrics can be visualized at the same time, their prediction files are kept separately under different folders.
- Classification metrics can be filtered according to the chosen classes.

For now, streamlit tab is created in `app/views/model_quality` tab, also classification performance metrics are also implemented here which is not ideal. However, there will be very fundamental changes in **Model quality** tab both in terms of reorganization of tabs and global filters. So, I haven't put too much effort to organizing and refactoring things that will already change very soon.


src/encord_active/lib/model_predictions/reader.py
```diff
@@ -23,13 +25,38 @@ class OntologyObjectJSON(TypedDict):
     color: str
 
 
+class OntologyClassificationJSON(TypedDict):
```
frederik-encord, line 28: If you like my other suggestion, this definition needs to go to the writer to avoid cyclic imports.

src/encord_active/app/views/model_quality.py
```diff
@@ -2,87 +2,258 @@
...
+                    num_samples = model_predictions.shape[0]
+
+                metric_columns = list(get_state().predictions.metric_datas_classification.predictions.keys())
+                metric_importance_chart = create_metric_importance_charts(
+                    model_predictions_matched_filtered,
+                    metric_columns=metric_columns,
+                    num_samples=num_samples,
+                    prediction_type=MainPredictionType.CLASSIFICATION,
+                )
+                st.altair_chart(metric_importance_chart, use_container_width=True)
```
frederik-encord, line 242: This needs to be wrapped in a try catch as it will raise an error if there are not enough examples.

frederik-encord (changes requested):
Issues that I found along the way were:
1. When I click the "Performance By  metrics", "False Positive", etc (not includint "Metrics") for a classification project, nothing happens?
2. Since the only thing that we need to do for existing projects (at least for object predictions) is to move the prediction files into an `objects` directory, perhaps we should do that automatically if we detect that there are model predictions available in the project? There may be some caveats around classifications stored as objects though? Don't think that any customer got that far though?

src/encord_active/app/views/model_quality.py
```diff
@@ -2,87 +2,258 @@
...
+                    return
+
+                if labels_all is None:
+                    st.error("Couldn't load labels properly")
+                    return
+
+                with sticky_header():
+                    common_settings_classifications()
+
+                model_predictions_matched = match_predictions_and_labels(model_predictions, labels_all)
```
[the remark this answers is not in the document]
Gorkem-Encord: Indeed it is quite light, it corresponds to `compute_mAP_and_mAR` function, which adds extra column that indicates the  TP status.

frederik-encord (changes requested):

frederik-encord (approved):

PR #240, opened 2023-03-10
fix (coco import): fix several COCO import bugs

Gorkem-Encord:


Gorkem-Encord:
@frederik-encord could you check this PR?

frederik-encord (approved):

PR #246, opened 2023-03-14
alexb/refresh data link for signed urls

alex-encord:
A small change: if a data link does not match the pattern for Encord hosting, refresh the data link.
This will ensure that signed URLs are refreshed for private cloud integrations.


src/encord_active/lib/project/project.py
```diff
@@ -253,13 +256,19 @@ def download_images_from_data_unit(lr: LabelRow, project_file_structure: Project
     lr_structure.images_dir.mkdir(parents=True, exist_ok=True)
     frame_pths: List[Path] = []
     data_units = sorted(lr.data_units.values(), key=lambda du: int(du["data_sequence"]))
+    # If customer uses private cloud integration with signed URLs, refresh data links for label row
+    pat = re.compile(r'^(?!https://storage.googleapis.com/cord-ai-platform.appspot.com/.*)')
+    if any([re.match(pat, du['data_link']) for du in data_units]):
+        lr = encord_project.get_label_row(label_hash)
+        data_units = sorted(lr.data_units.values(), key=lambda du: int(du["data_sequence"]))
```
frederik-encord, line 263: This seems to defeat the whole purpose of the `download_label_rows`? Now we're essentially downloading every label_row twice?

frederik-encord: .. for the cloud integration customers at least.

frederik-encord (dismissed):
I've approved but the re-download seems sub-optimal and we should very soon start refactoring this properly. 

.. Also, what if there are 100 data units? Then we'll only get the first few of them?

src/encord_active/lib/project/project.py
alex-encord: Yes exactly, this is generally quite a sub optimal fix on top of a sub optimal skeleton. I will make a follow up PR to clean up the entire download flow up as priority was to unblock the customer.

frederik-encord (approved):

PR #253, opened 2023-03-16
feat(use-state): make hooks truly generic

Encord-davids:
This change would allow creating states with a default of `None`, e.g. 

```python
state = UseState[Optional[int]]()

state.value == None # -> True, and the type of is correct
```

### DIFF
--- a/src/encord_active/app/data_quality/sub_pages/explorer.py
+++ b/src/encord_active/app/data_quality/sub_pages/explorer.py
@@ -28,7 +28,7 @@ from encord_active.app.common.components.tags.individual_tagging import multisel
 from encord_active.app.common.components.tags.tag_creator import tag_creator
 from encord_active.app.common.page import Page
 from encord_active.app.common.state import get_state
-from encord_active.app.common.state_hooks import use_state
+from encord_active.app.common.state_hooks import UseState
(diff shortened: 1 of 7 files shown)


src/encord_active/app/data_quality/sub_pages/explorer.py
```diff
@@ -160,14 +161,14 @@ def render_plotly_events(embedding_2d: DataFrame[Embedding2DSchema]) -> Optional
 
     new_selection = plotly_events(fig, click_event=False, select_event=True)
 
-    if new_selection != get_selection():
-        set_should_select(True)
-        set_selection(new_selection)
+    if new_selection != selection.value:
```
frederik-encord, line 164: So at this point, the `selection` is typed as a `List[dict]` but it's really `None` right?

frederik-encord: Perhaps we should type it `Optional[List[dict]]`

frederik-encord (dismissed):

src/encord_active/app/data_quality/sub_pages/explorer.py
Encord-davids: Absolutely 👍

frederik-encord (approved):

PR #260, opened 2023-03-20
fix: stop early with bad transformer file and raise error on module load

frederik-encord:


eloy-encord:
I don't have it clear if we fail gracefully in case the module doesn't have a python extension.

frederik-encord:
Good catch. Should be fixed now. Also fixed some cosmetics for the `ea metric add` command 💅🏻

PR #265, opened 2023-03-20
fix: remove message after interaction with page

frederik-encord:


Encord-davids (dismissed):

yogesh-encord:
I believe we can remove all the other set_updates([]) now, testing it out

yogesh-encord (approved):

PR #274, opened 2023-03-22
fix(paths): use ProjectFileStructure instead of raw paths

yogesh-encord:


eloy-encord:
@yogesh-encord Can you double check that the refactoring you did in the label transformers examples still allows the user to copy/paste them and they still work? I'm most worried about the relative import `examples.utils` you used there.

yogesh-encord:
> @yogesh-encord Can you double check that the refactoring you did in the label transformers examples still allows the user to copy/paste them and they still work? I'm most worried about the relative import `examples.utils` you used there.

Good point! Removed those changes, thanks!!

frederik-encord (approved):

PR #285, opened 2023-03-24
Docs: add active learning examples

eloy-encord:
Includes:
1. Facelift of the existing examples to make them more appealing to the end user.
2. Add train and test MNIST sandbox projects.
3. Minor improvements to the acquisition functions' model wrapper to fit to user needs (used in the examples).

### DIFF
--- a/src/encord_active/lib/common/active_learning.py
+++ b/src/encord_active/lib/common/active_learning.py
@@ -6,18 +6,18 @@ import pandas as pd
 
 from encord_active.lib.common.iterator import DatasetIterator
 from encord_active.lib.metrics.acquisition_functions import BaseModelWrapper
-from encord_active.lib.metrics.execute import execute_metrics
 from encord_active.lib.project.project_file_structure import ProjectFileStructure
 
 
-def get_data(project_fs: ProjectFileStructure, model: BaseModelWrapper, class_name: str, subset_size=None):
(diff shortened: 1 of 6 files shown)


src/encord_active/lib/common/active_learning.py
```diff
@@ -0,0 +1,91 @@
...
+def get_data_sample(project_fs: ProjectFileStructure, data_hash: tuple[str, str], class_name: str):
+    label_hash, du_hash = data_hash
+    lr_struct = project_fs.label_row_structure(label_hash)
+
+    # get classification label
+    label_row = json.loads(lr_struct.label_row_file.read_text())
+    class_label = get_classification_label(label_row, du_hash, class_name=class_name)
+
+    # get image path
+    image_path = lr_struct.images_dir / f"{du_hash}.{label_row['data_units'][du_hash]['data_type'].split('/')[-1]}"
```
frederik-encord, line 40:
```suggestion
    image_path = next((dus.path for dus in lr_struct.iter_data_unit()))
```
This avoids data misses - as mentioned, we should always use the structures now. No manual path creation. It will fail.

frederik-encord (changes requested):

src/encord_active/lib/common/active_learning.py
eloy-encord: fixed

PR #304, opened 2023-03-29
feat: smart refresh of project data

eloy-encord:
Includes:
1. Refresh project capabilities
2. CLI refresh command
3. Medium refactor of data and labels download capabilities to make an easier transition to the new-to-come filesystem

Missing:
1. Docs in the CLI (on it on the morning of March 30, 2023)

Problem: The mappings in the label rows structures are troublesome as they weren't designed having in mind that the data may change over time. They don't have a way to construct paths that don't exist yet so I had to manually force the name of the files in the filesystem in the project code (at least this is not a big issue as it was already done that way).

### DIFF
--- a/docs/docs/cli.md
+++ b/docs/docs/cli.md
@@ -15,12 +15,12 @@ Here is a list of all the top-level commands:
 
 ```
 quickstart         Start Encord Active straight away 🏃💨
-download           Download a sandbox dataset to get started. 📁
+download           Download a sandbox dataset to get started 📁
 init               Initialize a project from your local file system 🌱
 import             Import Projects or Predictions ⬇️
+refresh            Sync data and labels from a remote Encord project 🔄
 visualize          Launch the application with the provided project ✨
-metric             Manage project's metrics.
-metricize          Execute metrics on your data and predictions 🧠
+metric             Manage project metrics 📋
 print              Print useful information 🖨️
 config             Configure global settings 🔧
 ```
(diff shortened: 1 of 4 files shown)


src/encord_active/lib/project/project.py
```diff
@@ -92,6 +93,38 @@ def from_encord_project(self, encord_project: EncordProject) -> Project:
...
+        """
+
+        if not self.project_meta.get("has_remote", False):
+            raise AttributeError("The project does not have a remote project associated to it.")
+
+        project_hash = self.project_meta.get("project_hash")
+        if project_hash is None:
+            raise AttributeError(
+                "The project metadata is missing the reference to the remote project in Encord Annotate. "
+                "Add the project hash of the remote project to the property `project_hash` in the file "
```
frederik-encord, line 110:
```suggestion
                "The project is not associated with a remote project to refresh from. Aborting."
```

src/encord_active/lib/project/project.py
```diff
@@ -92,6 +93,38 @@ def from_encord_project(self, encord_project: EncordProject) -> Project:
...
+                f"{self.file_structure.project_meta}."
+            )
+
+        ssh_key_path = self.project_meta.get("ssh_key_path")
+        if ssh_key_path is None:
+            raise AttributeError(
+                "The project metadata is missing the path to the private SSH key needed to log into Encord Annotate. "
+                f"Add such path to the property `ssh_key_path` in the file {self.file_structure.project_meta}."
+            )
```
frederik-encord, line 119: There is a utility function somewhere to both look for the ssh key in the project meta and in the EA settings.

frederik-encord (changes requested):

docs/docs/cli.md
```diff
@@ -231,9 +231,33 @@ Options:
...
+2. The hash of the remote Encord project.
+3. The path to the private Encord user SSH key.
+
+Use this snippet to fill the required fields and remember to substitute `project_hash` and `ssh_key_path` attributes with yours.
+```yaml
+has_remote: true
+project_hash: remote-encord-project-hash
+ssh_key_path: private-encord-user-ssh-key-path
+```
+
```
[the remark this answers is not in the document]
eloy-encord: yeah, I failed this one being way too explicit. There is really no need for such low level information that will raise more questions than answers. I'm going to use your suggestion as ground truth here. Thanks!

src/encord_active/lib/project/project.py
eloy-encord: while I didn't find the right utility function I added the same behaviour in the `dict.get()` :+1:

frederik-encord (approved):

PR #312, opened 2023-04-03
fix: moves pagination before gallery

Encord-davids:


frederik-encord (changes requested):
I think that we should keep the slider in the bottom of the page. 

[here](https://github.com/encord-team/encord-active/blob/46bf854f3b292268ac22791b5cd0e97a1c8b2c85/src/encord_active/app/data_quality/sub_pages/explorer.py#L224-L254) is an example of how we can do it.

Encord-davids:
> I think that we should keep the slider in the bottom of the page.
> 
> [here](https://github.com/encord-team/encord-active/blob/46bf854f3b292268ac22791b5cd0e97a1c8b2c85/src/encord_active/app/data_quality/sub_pages/explorer.py#L224-L254) is an example of how we can do it.

I was going to move the slider to the new explorer so didn't really spend time solving this, but thanks, I'll copy that 🙏

PR #313, opened 2023-04-04
Fix (active learning): extending to radio group classifications

Gorkem-Encord:


src/encord_active/lib/common/active_learning.py
```diff
@@ -49,8 +49,13 @@ def get_classification_label(label_row, du_hash: str, class_name: str):
     if len(filtered_class) == 0:
         return None
     class_hash = filtered_class[0]["classificationHash"]
-    class_label = label_row["classification_answers"][class_hash]["classifications"][0]["answers"]
-    return class_label
+    answers = label_row["classification_answers"][class_hash]["classifications"][0]["answers"]
```
frederik-encord, line 52:
```suggestion
    answers = label_row["classification_answers"].get(class_hash, {}).get("classifications", [{}])[0].get("answers")
```

frederik-encord: Something like that should make the code more safe.

frederik-encord (changes requested):

src/encord_active/lib/common/active_learning.py
Gorkem-Encord: very nice catch, we fall into that trap quite often :)

frederik-encord (approved):

PR #326, opened 2023-04-12
feat: add premium feature support

frederik-encord:


src/encord_active/app/common/components/assistant.py
```diff
@@ -0,0 +1,135 @@
...
+        filter_merged_metrics_by_query_result(result)
+
+    return query_with_embedding
+
+
+assistant_search_mode = AssistantMode(
+    "Data Search", make_query_with_embedding(get_state().querier.search_with_metric_data, TextQuery)
+)
+assistant_clip_mode = AssistantMode(
+    "CLIP search", make_query_with_embedding(get_state().querier.search_with_clip, CLIPQuery)
```
Encord-davids, line 100:
```suggestion
    "Search", make_query_with_embedding(get_state().querier.search_with_clip, CLIPQuery)
```

src/encord_active/lib/premium/querier.py
```diff
@@ -0,0 +1,64 @@
...
+        return response.json()
+
+    @property
+    def premium_available(self) -> bool:
+        if self._premium_available is None:
+            try:
+                self._premium_available = bool(self.execute(timeout=0.5))
+            except (ConnectionError, ConnectionRefusedError, Exception):
+                pass
+        return self._premium_available or False
```
Encord-davids, line 45: a `0.5` timeout seems reasonable in a `localhost` setup, but on the network that might be too short.

May I suggest that we rely purely on the `ENV` variable, i.e. premium is available only if env var is present, in which case we raise an error if the timeout is met. (maybe half a minute?)

Encord-davids (dismissed):

PR #328, opened 2023-04-17
refactor: detach data units from label rows in the project file structure

eloy-encord:
Objectives:
1. Detach data units from label rows in the logic to be able to fetch unannotated data without modifying label row states.
2. Move data units relevant info (not its labels) to the db while not breaking older projects.
3. Refactor `ProjectFileStructure` to enable calls from it to the db.
4. Update `DBConnection` to accept different project file structures, not only the default one, in case the user decides to store its data in some other layout (or we change the way the project is structured).


eloy-encord:
Closing this draft in favour of https://github.com/encord-team/encord-active/pull/329 because this one pulled already non-existent (squashed) commits.

PR #349, opened 2023-04-27
fix: encord sdk bad datetime format

eloy-encord:
Encord SDK is sending a bad format for the datetimes with comma valued datetimes like the following `2022-08-10 17:34:58.649359` and it affects the refresh capability as this error only occurs at label row metadata level, not the label row itself. Now, we trim the faulty numbers and leave it in the proper format while the Encord SDK push a fix and new version.

### DIFF
--- a/src/encord_active/lib/project/project.py
+++ b/src/encord_active/lib/project/project.py
@@ -166,6 +166,10 @@ class Project:
             for lrm in encord_project.list_label_rows()
             if lrm.label_hash is not None
         }
+        for meta in label_row_meta.values():
+            meta["created_at"] = meta["created_at"].rsplit(".", maxsplit=1)[0]
+            meta["last_edited_at"] = meta["last_edited_at"].rsplit(".", maxsplit=1)[0]
+
         self.file_structure.label_row_meta.write_text(json.dumps(label_row_meta, indent=2), encoding="utf-8")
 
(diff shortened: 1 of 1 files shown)


frederik-encord:
I am not sure why clipping off the microseconds would fix the issue? The ISO format of the timestamp should be correct no?

eloy-encord:
We save label rows' metadata with the method `encord_project.list_label_rows()` while we use `encord_project.label_rows()` in other places as it guarantees that we only fetch annotated data because of the must exist `label_hash` restriction. Problem is that `list_label_rows()` sends dates with the trailing `.000000`, where `label_rows()` doesn't which creates the difference when comparing the supposedly same date. This event causes that all label rows are refreshed in each refresh call since the difference between stored date and new date is the same.

We could use `label_rows()` instead of `list_label_rows()` for saving but I'm not aware if that's a good idea as the later was supposed an improved version of the first and could cause some unexpected disruption with other metadata fields. Upgrading the Encord SDK version still doesn't solve the issue as `list_label_rowsV2()` uses the same endpoint as the first version and the dates are also corrupted. This problem should go away with a fix on Encord SDK or when we start fetching all the data, including those with no label hash, as we will use only one method of the SDK and if gets corrupted then it will sort itself out (corrupted == corrupted).

frederik-encord (approved):

frederik-encord:
Now I understand - approved

PR #352, opened 2023-04-27
Feat (project diff): project metrics difference

Gorkem-Encord:


src/encord_active/app/project_comparison/project_similarity.py
```diff
@@ -0,0 +1,165 @@
...
+
+def calculate_metric_similarity(array_1: np.ndarray, array_2: np.ndarray) -> float:
+    ks_score, _ = ks_2samp(array_1, array_2)
+    return 1 - ks_score
+
+
+def render_2d_metric_similarity_container(
+    all_metrics: list, merged_metrics_1: pd.DataFrame, merged_metrics_2: pd.DataFrame, project_name_2: str
+):
+    metrics_filtered = []
```
frederik-encord, line 38:
```suggestion
    metrics_filtered: List[str] = []
```

src/encord_active/app/project_comparison/project_similarity.py
```diff
@@ -0,0 +1,165 @@
...
+        return
+
+    project_values_2 = merged_metrics_2[[metric_name_1, metric_name_2]].copy().dropna()
+    project_values_2["project"] = project_name_2
+
+    # TODO change append to concat
+    project_values = pd.concat([project_values_1, project_values_2], ignore_index=True)
+
+    fig = render_2d_metric_similarity_plot(
+        project_values, cast(str, metric_name_1), cast(str, metric_name_2), project_name_1, project_name_2
```
frederik-encord, line 103: If we type the metric lists above, we won't have to cast anything here.
Generally, when we start casting something, it should raise an internal alarm. It shouldn't be necessary.

src/encord_active/app/project_comparison/project_similarity.py
```diff
@@ -0,0 +1,165 @@
...
+
+    for metric_name in metrics_filtered:
+        if metric_name == metric_name_1:
+            continue
+        metric_df_1 = merged_metrics_1[metric_name_1].dropna().to_frame()
+        metric_df_2 = merged_metrics_1[metric_name].dropna().to_frame()
+
+        merged_metric = metric_df_1.join(metric_df_2, how="inner")
+        if merged_metric.shape[0] > 0:
+            options_for_second_metric.append(metric_name)
```
frederik-encord, line 67: Even after fixing the typo below, I don't understand the purpose of this piece of code. For two different projects, the merging of the identifiers should always have shape `[0, 2]` because they don't share label hashes. So the result will always be an empty list.

```suggestion
    metric_df_1 = merged_metrics_1[metric_name_1].dropna().to_frame()
    for metric_name in metrics_filtered:
        if metric_name == metric_name_1:
            continue
        metric_df_2 = merged_metrics_2[metric_name].dropna().to_frame()

        merged_metric = metric_df_1.join(metric_df_2, how="inner")
        if merged_metric.shape[0] > 0:
            options_for_second_metric.append(metric_name)
```

src/encord_active/app/project_comparison/project_similarity.py
```diff
@@ -0,0 +1,165 @@
...
+        merged_metrics_2 = MergedMetrics().all()
+        project_name_2 = project_metas[selected_project]["project_title"]
+        DBConnection.set_project_file_structure(ProjectFileStructure(get_state().project_paths.project_dir))
+
+        all_metrics = sorted([metric for metric in fetch_metrics_meta(get_state().project_paths)])
+
+        metric_similarities = {"metric": [], "similarity_score": []}
+        for metric in all_metrics:
+            if metric in METRICS_TO_EXCLUDE:
+                continue
```
frederik-encord, line 134: Is it on purpose that we don't remove the metric from `all_metrics`?
The consequence is that the 2D plot will still allow you to choose, e.g., random as one of the axes.

src/encord_active/app/project_comparison/project_similarity.py
```diff
@@ -0,0 +1,165 @@
...
+
+    if len(metrics_filtered) <= 1:
+        st.write("At least two metrics should be generated for 2D metric similarity plot")
+        return
+
+    metric_selection_col_1, metric_selection_col_2 = st.columns(2)
+    metric_name_1 = metric_selection_col_1.selectbox(
+        "Select the first metric",
+        options=metrics_filtered,
+        index=0,
```
frederik-encord, line 51: Would be nice to default to the two least similar metrics such that the plot looks most interesting from the beginning.

src/encord_active/lib/charts/project_similarity.py
```diff
@@ -0,0 +1,57 @@
...
+
+def render_2d_metric_similarity_plot(
+    project_values: pd.DataFrame, metric_name_1: str, metric_name_2: str, project_name_1: str, project_name_2: str
+):
+    fig = px.scatter(
+        project_values,
+        x=metric_name_1,
+        y=metric_name_2,
+        color="project",
+        color_discrete_map={project_name_1: "#5658dd", project_name_2: "red"},  # Encord brand colors: #5658dd, #89e7b6
```
frederik-encord, line 39: We could use color definitions defined in the project
So import
```python
from encord_active.lib.common.colors import Color
```
And then you can do
```suggestion
        color_discrete_map={
            project_name_1: Color.PURPLE.value,
            project_name_2: Color.ORANGE.value,
        },
```

frederik-encord (changes requested):
Overall a very nice contribution! I left a handful of comments to improve the code quality slightly.

[💅 Nitpick]
We should specify in `fetch_metrics_meta` that it returns `Dict[str, Dict[Any]]` to make it a bit more explicit what we're working with here.

src/encord_active/app/project_comparison/project_similarity.py
Gorkem-Encord: 👍 Put Brightness vs. contrast for now. They both seem good. Unbounded metrics like Area etc. seem not attractive I think.

src/encord_active/app/project_comparison/project_similarity.py
Gorkem-Encord: Actually, it is not a typo; it was intentional. The aim here is to find compatible metrics with each other that can be drawn on 2D plot. It iterates through metrics and sees if its identifier is compatible to the chosen one.

src/encord_active/app/project_comparison/project_similarity.py
Gorkem-Encord: You are certainly right about casting, however what happens in this case is that when streamlit returns the selected name as below:

```
    metric_name_1 = metric_selection_col_1.selectbox(
        "Select the first metric",
        options=metrics_filtered,
        index=metrics_filtered.index("Brightness") if "Brightness" in metrics_filtered else 0,
        key=f"project_comparison_metric_selection_1_{get_state().project_paths.project_dir.name}_{project_name_2}",
    )
```

the `metric_name_1` is of type `Optional[str]`; therefore mypy raises an error in this line, that is why I had to restrict its type. Another solution might be checking the type as below just before line 103?

```
instance(metric_name_1, str)
instance(metric_name_2, str)
```

What do you think?

src/encord_active/app/project_comparison/project_similarity.py
Gorkem-Encord: Yeah, it is on purpose. These excluded metrics would affect the average similarity score in not a meaningful way so I excluded them but thought the user might still want to see their distributions.

src/encord_active/app/project_comparison/project_similarity.py
frederik-encord: Checking for none is what we should do. It's the safest.

src/encord_active/app/project_comparison/project_similarity.py
frederik-encord: See line 71

src/encord_active/app/project_comparison/project_similarity.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # Second selectbox should be populated according to the selected metric in the first one
+    # due to different metric types. E.g., 'Area' is not compatible with 'object aspect ratio'
+    options_for_second_metric = []
+
+    metric_df_1 = merged_metrics_1[metric_name_1].dropna().to_frame()
+    for metric_name in metrics_filtered:
+        if metric_name == metric_name_1:
+            continue
+        metric_df_2 = merged_metrics_1[metric_name].dropna().to_frame()
```
frederik-encord, line 63: This line is the typo that I meant.

frederik-encord (approved):

PR #357, opened 2023-05-02
fix: revert in-repo prisma and fix client generation and schema migrations

Encord-davids:


src/encord_active/lib/db/prisma_init.py
```diff
@@ -0,0 +1,28 @@
...
+prisma_run = silence_stdout(run)
+
+GENERATED_PRISMA_SCHEMA_FILE = Path(prisma_module_file).parent / "schema.prisma"
+PRISMA_SCHEMA_FILE = Path(__file__).parent / "schema.prisma"
+
+
+def ensure_prisma_db(db_path: Path):
+    env = {"MY_DATABASE_URL": f"file:{db_path}"}
+    failure = prisma_run(["migrate", "deploy", f"--schema={PRISMA_SCHEMA_FILE}"], env=env)
+    # TODO: remove me before release
```
eloy-encord, line 18: what about this todo? is ok to go with it to main?

src/encord_active/lib/db/connection.py
```diff
@@ -1,13 +1,23 @@
...
-from pathlib import Path
 from typing import Optional
 
-from encord_active.lib.db.__autogen__.cli.prisma import run
-from encord_active.lib.db.__autogen__.client import Prisma
-from encord_active.lib.db.__autogen__.types import DatasourceOverride
-from encord_active.lib.file_structure.base import BaseProjectFileStructure
+from encord_active.lib.db.prisma_init import generate_prisma_client
+
+try:
```
eloy-encord, line 6: should we move the try/except to `prisma_init` module so after calling that module we can ensure that prisma effectively works?

src/encord_active/lib/project/project_file_structure.py
```diff
@@ -2,7 +2,8 @@
 from pathlib import Path
 from typing import Iterator, NamedTuple, Optional, Union
 
-from encord_active.lib.db.__autogen__ import models
+from prisma import models
```
eloy-encord, line 5: probably we should add prisma initialization before importing the models, or moving this import to after the prisma db connection import

src/encord_active/lib/file_structure/base.py
```diff
@@ -2,7 +2,8 @@
 from pathlib import Path
 from typing import Iterator, Optional, Union
 
-from encord_active.lib.db.__autogen__ import models
+from prisma import models
```
eloy-encord, line 5: probably we should add prisma initialization before importing the models

eloy-encord (changes requested):

src/encord_active/lib/db/prisma_init.py
Encord-davids: we want it to go to main for a short while so everyone who already used prisma can start all projects and reset them. after that we can remove

src/encord_active/lib/project/project_file_structure.py
Encord-davids: Good point, I'll move the try/except block to the `db` module init

frederik-encord (approved):

eloy-encord (approved):

PR #371, opened 2023-05-04
feat: enable exported projects to work properly with prisma db

eloy-encord:
Changes include:
1. Addition of `DataHashMapping` class to ease the overall process of storing mappings while allowing easy comprehension on what's being done (avoid passing dicts every time).
2. Change db's data hashes for the new ones that result from the `Export to Encord` feature.
3. Enable the `Export to Encord` feature.


eloy-encord:
Added a few commits tackling dirty cache and cache misses problems after the data hashes change as a result of exporting a project to Encord Annotate.

frederik-encord (approved):

PR #379, opened 2023-05-05
docs: minor fixes and updates

eloy-encord:


docs/docs/import/import-coco-project.md
```diff
@@ -60,6 +60,6 @@ When the process is done, follow the printed instructions to open the app or see
 
 :::info
 
-If you want the dataset to be available on the Encord platform, please refer to the [Export documentation](../user-guide/filter_export#export-to-encord).
```
frederik-encord, line 63: think that "the" should actually be there.

docs/docs/index.mdx
```diff
@@ -68,27 +68,30 @@ import Link from "@docusaurus/Link";
...
+      <img
+        style={{ marginRight: 5 }}
+        alt="Downloads"
+        src="https://static.pepy.tech/badge/encord-active"
+      />
     </td>
   </tr>
 </table>
 
-## ❓ What is Encord Active?
```
frederik-encord, line 75: I don't think that we should remove this title.

docs/docs/sdk/prediction-import/index.md
```diff
@@ -23,7 +23,7 @@ from encord_active.lib.project import Project
 
 project_path = Path("/path/to/your/project/root")
 
-predictions: List[Prediction] = ...  # Your list of predictions
```
frederik-encord, line 26: I would keep `List` and then add 
```python
from typing import List
```
in the top of the block.

frederik-encord (changes requested):

docs/docs/sdk/prediction-import/index.md
eloy-encord: We can use safely use `list` because `typing.List` (and a few others hints from that package) are deprecated since 3.9 and `list` is now the recommended type hint.

docs/docs/index.mdx
eloy-encord: I removed that line to increase the consistency degree between Encord Annotate and Active docs. Check [annotate](https://docs.encord.com/) against [active](https://docs.encord.com/active/docs). Let me know what you think afterwards.
I prefer to keep that line in the github repo readme but doesn't seem necessary for the docs.

docs/docs/sdk/prediction-import/index.md
frederik-encord: Hmm okay - we use `List` everywhere in the code base, which is why I suggested it. Anyway, it's fine then.

frederik-encord (approved):

PR #381, opened 2023-05-08
docs: update code/text/links in 'import KITTI predictions' section

eloy-encord:


docs/docs/sdk/prediction-import/kitti.mdx
```diff
@@ -3,23 +3,22 @@ title: KITTI Bounding Boxes
...
 ---
 
-:::caution
-
-This works for bounding boxes only.
-
-:::
+The KITTI dataset is a widely-used computer vision dataset for benchmarking and evaluating autonomous driving systems.
+It consists of a collection of high-resolution images, lidar data, and camera calibration parameters captured from a moving platform driving through urban environments.
+The dataset includes a variety of tasks such as object detection, tracking, 3D object detection, and semantic segmentation.
```
frederik-encord, line 8: What dataset? We only care about the format here I think. Otherwise, perhaps say 
```suggestion
The [original dataset](https://www.cvlibs.net/datasets/kitti/) includes a variety of tasks such as object detection, tracking, 3D object detection, and semantic segmentation.
```

docs/docs/sdk/prediction-import/kitti.mdx
```diff
@@ -32,14 +31,14 @@ labels_root
 
 That is, a root directory with two components:
 
-1. A subdirectory named "labels" that contains text files with names that start with the `data_hash` followed by two underscores
-2. A json file which maps class names to Encord ontology classes
+1. A subdirectory named "labels" that contains TXT/CSV files, with filenames that begin with the `data_hash` followed by two underscores.
```
frederik-encord, line 34:
```suggestion
1. A subdirectory named "labels" that contains TXT/CSV files with filenames that begin with the `data_hash` followed by two underscores.
```

docs/docs/sdk/prediction-import/kitti.mdx
```diff
@@ -83,46 +86,57 @@ The JSON class map needs to follow the following structure:
...
 
 client = EncordUserClient.create_with_ssh_private_key(private_key)
-project = client.get_project(project_hash=meta["project_hash"])
+encord_project = client.get_project(project_hash=project_meta["project_hash"])
 
-# NB: Remember to include the first code snippet on this page.
-print({o["featureNodeHash"]: o["name"] for o in project.ontology["objects"]})
-# Outputs somtihing similar to
+print({o["featureNodeHash"]: o["name"] for o in encord_project.ontology["objects"]})
+# Example output:
```
frederik-encord, line 107: Since we're updating this anyway - it would probably be better to just load the local `ontology.json` file instead and use that. That would remove the requirement of the project being uploaded to Annotate.

frederik-encord (changes requested):

docs/docs/sdk/prediction-import/kitti.mdx
eloy-encord: Just giving an introduction for those that are looking for ways to import their data so they know it's not a fit.

docs/docs/sdk/prediction-import/kitti.mdx
eloy-encord: purfect observation 👍

docs/docs/sdk/prediction-import/kitti.mdx
eloy-encord: I rewrote that section to actually refer to the label format instead of the dataset.

frederik-encord (approved):

PR #382, opened 2023-05-09
chore: revert dockerfile

Encord-davids:


Dockerfile
```diff
@@ -2,55 +2,31 @@ FROM python:3.9-slim
...
 
-COPY run.sh /home/appuser
-
-USER root
-RUN mkdir /data && chown appuser: /data -R
-USER appuser
+ENV STREAMLIT_SERVER_ADDRESS=0.0.0.0
 
-RUN git config --global --add safe.directory '*'
-EXPOSE 8000
```
frederik-encord, line 54: I think that we would still want to expose port 8000 no?

Encord-davids: Don't know if needed but lets add it 👍

frederik-encord (approved):

PR #399, opened 2023-05-11
feat: clip embeddings

frederik-encord:
Gets rid of EfficientNet and introduces CLIP as the model for embedding images.

A couple of additional improvements are:
1. The embedder is now a class so no need to pass transforms around
2. When embedding images, we now batch them + identify the maximal batch size for the GPU to be able to handle the load. This alone yields a +10x speedup on embedding images.
3. PyTorch has been upgraded to 2.0.1


Encord-davids (approved):

frederik-encord:
@Encord-davids I've updated this. I had to introduce a larger refactor to avoid circular imports. I basically had to move the different metric types (AnnotationType, DataType, etc.) to it's own file. It shouldn't be a breaking change as we will still be able to do `from encord_active.lib.metrics.metric import DataType`. The only one that is a breaking change is AnnotationType` which is no longer imported in `metrics/metric.py`. Could import it just for backward compatibility?

Encord-davids (approved):

PR #401, opened 2023-05-12
feat: Move image loading to earlier, out of embeddings / metrics.

james-encord:
First part of work on removing need to download images & the file structure as source of truth.

This moves image opening logic to earlier, so it can be more easily changed in follow-up PRs.

### DIFF
--- a/src/encord_active/lib/embeddings/embeddings.py
+++ b/src/encord_active/lib/embeddings/embeddings.py
@@ -45,7 +45,7 @@ def assemble_object_batch(data_unit: dict, image: Image.Image) -> List[Image.Ima
                     continue
                 x, y, w, h = out
 
-                img_patch = image[y: y + h, x: x + w]
+                img_patch = image[y : y + h, x : x + w]
                 img_batch.append(Image.fromarray(img_patch))
             except Exception as e:
                 logger.warning(f"Error with object {obj['objectHash']}: {e}")


Encord-davids:
Great work! 
Could you just change the PR title to follow our [convention](https://docs.encord.com/active/docs/contributing#commit-convention)?🙏

src/encord_active/lib/common/utils.py
```diff
@@ -83,17 +84,12 @@ def load_json(json_file: Path) -> Optional[dict]:
...
 
-    image_corrupted = False
-    if img_pth is not None and img_pth.is_file():
-        try:
-            image = cv2.imread(img_pth.as_posix())
-            return image.shape[:2]
-        except Exception:
-            image_corrupted = True
+    if image is not None:
+        return image.size
```
frederik-encord, line 92: Pillow's `Image.size` is a tuple `(widht, height)` which is the oposite order of what's returned elsewhere.

frederik-encord (commented):
Looking really good - a couple of things that we need to make sure to do is:
1. Update documentation around custom metrics - to use images instead of paths
2. Update example notebook that defines custom metric
3. See if there are any other places where we use the `iterator.iterate` in documentation or code and update that as well.
4. Write out in Discord that this will change with the next release.

src/encord_active/lib/common/iterator.py
```diff
@@ -79,9 +81,15 @@ def iterate(self, desc: str = "") -> Generator[Tuple[dict, Optional[Path]], None
...
-                            self.project_file_structure.label_row_structure(label_hash).iter_data_unit(self.du_hash)
-                        ).path
+                        img_path = next(
+                            self.project_file_structure.label_row_structure(label_hash).iter_data_unit(self.du_hash),
+                            None,
+                        )
+                        if img_path is not None:
+                            image = Image.open(img_path.path)
+                        else:
+                            image = download_image(data_unit["data_link"])
```
frederik-encord, line 91: This one can be dangerous. These urls have TTL so might be invalid at the time we try to fetch them here.

src/encord_active/lib/common/utils.py
```diff
@@ -438,6 +434,22 @@ def download_file(
...
+        raise ConnectionError(f"Something happened, couldn't download file from: {url}")
+
+    return Image.open(r.content)
+
+
+def convert_image_to_cv2(image: Image.Image) -> np.ndarray:
+    rgb_image = image.convert("RGB")
+    np_image = np.array(rgb_image)
+    ocv_image = np_image[:, :, ::-1].copy()
+    return ocv_image
```
frederik-encord, line 450: I would probably call this `convert_to_bgr` instead.
```suggestion
def convert_image_bgr(image: Image.Image) -> np.ndarray:
    rgb_image = image.convert("RGB")
    np_image = np.array(rgb_image)
    return np_image[:, :, ::-1].copy()
```

frederik-encord: ☝🏻 Remember to update usages if you change the name ;-)

src/encord_active/lib/model_predictions/iterator.py
```diff
@@ -203,7 +204,7 @@ def iterate(self, desc: str = "") -> Generator[Tuple[dict, Optional[Path]], None
 
                 du["labels"] = {"objects": objects, "classifications": classifications}
                 pth = self.get_image_path(fr_preds.iloc[0])
-                yield du, pth
+                yield du, Image.open(pth)
```
frederik-encord, line 207: Not sure if we want a try/except here to avoid crashing the iterator?

src/encord_active/lib/common/iterator.py
```diff
@@ -79,9 +81,15 @@ def iterate(self, desc: str = "") -> Generator[Tuple[dict, Optional[Path]], None
...
                     try:
-                        yield data_unit, next(
-                            self.project_file_structure.label_row_structure(label_hash).iter_data_unit(self.du_hash)
-                        ).path
+                        img_path = next(
+                            self.project_file_structure.label_row_structure(label_hash).iter_data_unit(self.du_hash),
+                            None,
+                        )
+                        if img_path is not None:
+                            image = Image.open(img_path.path)
```
frederik-encord, line 89: perhaps a try/except here if images are, e.g., only half way downloaded or corrupted in some other way.

frederik-encord (changes requested):

frederik-encord (dismissed):

frederik-encord (approved):

PR #413, opened 2023-05-17
feat: add confirmation button in the relabel feature

eloy-encord:


src/encord_active/app/actions_page/export_filter.py
```diff
@@ -580,32 +582,52 @@ def render_relabel_button(
...
+    for index, label_row in enumerate(encord_project.list_label_rows_v2(label_hashes=unique_lr_hashes), start=1):
+        label_row.workflow_reopen()
+        update_progress_bar(index / len(unique_lr_hashes))
+
+    clear_progress_bar()
+    label.empty()
+    label.text(
+        f"Data successfully submitted for labeling in Encord Annotate.\n"
+        f"A total of {len(unique_lr_hashes)} tasks have been assigned to the first labeling stage."
+    )
```
frederik-encord, line 629: Think that we should add a link to the project here as well for people to jump directly to the project.

eloy-encord: great idea!

frederik-encord (approved):

PR #425, opened 2023-05-22
feat: No longer always store state on the disk, instead support dynamic download of data from s3/external-storage for encord projects.

james-encord:
This is really not as clean as I'd like it do be, but without significant changes to the codebase its as clean as I could get it. As image loading is done in lot of different places in the codebase, and video handling is not as easy to abstract away when frames are referenced and we previously assumed the frame existed on file.

Tested for a variety of workloads locally, please could you locally checkout this branch and report any bugs as part of the review process as this adds a significant number of changes across the entire codebase.

One known issue: there is an api in server that attempts to always return a url for an image, this with changes will always break for videos as we no longer have frame extraction stored on file. I have set it to return None in this case, potential fix i'm looking at is creating a temporary link somehow served from the server that has a timeout.


src/encord_active/lib/db/schema.prisma
```diff
@@ -49,6 +51,6 @@ model LabelRow {
   data_type      String
   created_at     String
   last_edited_at String
-  location       String
   data_units     DataUnit[]
+  label_row_json String     @default("missing")
```
Encord-davids, line 55: I think making the field nullable is better than putting a string that isn't a valid json in a field that holds a json string.

```suggestion
  label_row_json String?
```

src/encord_active/lib/project/project_file_structure.py
```diff
@@ -31,116 +35,229 @@
...
     @property
-    def label_row_file(self) -> Path:
-        return self.path / "label_row.json"
+    def label_row_json(self) -> Dict[str, Any]:
+        with PrismaConnection(self._project) as conn:
+            entry = conn.labelrow.find_unique(where={"label_hash": self._label_hash})
+            label_row_json = "missing"
+            if entry is not None:
+                label_row_json = entry.label_row_json
+            return json.loads(label_row_json)
```
Encord-davids, line 74: [nitpick] 💅

This block will always throw an exception when `entry is not None` since `"missing"` isn't a valid json 🙈

frontend_components/src/encord_active_components/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -568,15 +570,36 @@ const ImageWithPolygons = ({
         shape,
       }))
     );
-  }, [image.current?.clientWidth, image.current?.clientHeight, item.id]);
-
+  }, [media.current?.clientWidth, media.current?.clientHeight, item.id]);
+  console.log('item', item)
+  console.log('check', item.video_timestamp !== null && item.video_timestamp !== undefined)
```
Encord-davids, line 575: no need for logs here

```suggestion
```

frontend_components/src/encord_active_components/frontend/src/components/explorer/api.ts
```diff
@@ -42,6 +42,7 @@ export const GroupedTagsSchema = z.object({
 export const ItemSchema = z.object({
   id: z.string(),
   url: z.string(),
+  video_timestamp: z.number().nullish(),
```
Encord-davids, line 45: [nitpick] 💅

should probably be consistent with the rest of the fields

```suggestion
  videoTimestamp: z.number().nullish(),
```

frontend_components/src/encord_active_components/frontend/src/components/explorer/api.ts
```diff
@@ -140,7 +141,7 @@ export const getApi = (
           `${baseUrl}/projects/${projectName}/items/${encodeURIComponent(id)}`
         )
       ).json();
-      return ItemSchema.parse({ ...item, url: `${baseUrl}/${url}` }) as Item;
+      return ItemSchema.parse({ ...item, url }) as Item;
```
Encord-davids, line 144: [nitpick] 💅

Since we aren't transforming the URL we don't need to destructure it

```suggestion
      return ItemSchema.parse(item);
```

and also the statement above to 
```ts
const item = await ...
```

(would have made as part of the previous suggestion but GH is not allowing suggestions on lines that aren't in the diff)

frederik-encord (dismissed):

PR #429, opened 2023-05-23
Feat (model performance): Make more intuitive 2D embedding plots

Gorkem-Encord:
For objects (bbox/polygons) we show the performance of each image based on TP/(TP+FP+FN). 

image


src/encord_active/lib/embeddings/utils.py
```diff
@@ -102,6 +102,10 @@ class Embedding2DSchema(IdentifierSchema, PointSchema2D):
     label: Series[str]
 
 
+class Embedding2DScoreSchema(Embedding2DSchema):
+    score: Series[float]
+
+
```
frederik-encord, line 108: Depending on whether this PR or #433 gets merged first, we'll have to move this definition to `src/encord_active/lib/embeddings/types.py`.

frederik-encord (approved):

src/encord_active/lib/embeddings/utils.py
Gorkem-Encord: You mean all the schemas in this file right

frederik-encord (approved):

PR #434, opened 2023-05-25
Feat (examples): get predictions using micro-models

Gorkem-Encord:


Gorkem-Encord:
These scripts are translated into jupyter notebooks and moved to https://github.com/encord-team/encord-notebooks.

PR #437, opened 2023-05-26
feat: adds data migrations and runs them on ensure project

Encord-davids:


src/encord_active/lib/db/data/migrations/202305261617_backfill_label_rows.py
```diff
@@ -0,0 +1,161 @@
...
+        old_style_data = conn.labelrow.find_first(where={"label_row_json": "missing"}) is not None or fill_label_rows
+        if not (fill_label_rows or fill_data_units or old_style_data):
+            return
+        migration_counter = tqdm(
+            None,
+            desc=f"Migrating Database: {pfs.project_dir}",
+        )
+        with conn.batch_() as batcher:
+            migration_counter.update(1)
+            for label_row in pfs.iter_labels():
```
james-encord, line 35: Need to re-add the iterdir fallback (secret_fallback_iter_files), otherwise migrate will fail if the database is completely empty.

james-encord (changes requested):

james-encord (approved):

PR #442, opened 2023-05-31
fix: syncing embeddings on subset export

frederik-encord:


src/encord_active/lib/encord/project_sync.py
```diff
@@ -140,9 +141,15 @@ def create_filtered_embeddings(
             continue
         embeddings = pickle.loads(curr_embedding_file.read_bytes())
         embeddings_df = pd.DataFrame.from_dict(embeddings)
-        embeddings_df = embeddings_df[embeddings_df["identifier"].isin(filtered_df.identifier)]
-        filtered_embeddings = embeddings_df.to_dict(orient="list")
+        if embedding_type == EmbeddingType.IMAGE:
+            label_hashes = filtered_df.identifier.str.split("_").str[0]
+            embeddings_df = embeddings_df[embeddings_df["label_row"].isin(label_hashes)]
```
Gorkem-Encord, line 146: This does not seem like to cover the image groups where label_hash are the same for the different data_units

src/encord_active/lib/encord/project_sync.py
```diff
@@ -140,9 +141,15 @@ def create_filtered_embeddings(
             continue
         embeddings = pickle.loads(curr_embedding_file.read_bytes())
         embeddings_df = pd.DataFrame.from_dict(embeddings)
-        embeddings_df = embeddings_df[embeddings_df["identifier"].isin(filtered_df.identifier)]
-        filtered_embeddings = embeddings_df.to_dict(orient="list")
+        if embedding_type == EmbeddingType.IMAGE:
+            label_hashes = filtered_df.identifier.str.split("_").str[0]
```
Gorkem-Encord, line 145: Would rename to `label_row_hash` to not to confuse the label hashes in the else block.

src/encord_active/lib/encord/project_sync.py
```diff
@@ -140,9 +141,15 @@ def create_filtered_embeddings(
...
+            embeddings_df = embeddings_df[embeddings_df["label_row"].isin(label_hashes)]
+        else:
+            label_hashes = filtered_df.identifier.str.split("_").str.get(3).dropna()
+            embeddings_df = embeddings_df[embeddings_df["labelHash"].isin(label_hashes)]
```
Gorkem-Encord, line 149: For the classification embeddings, `labelHash` refers to `classificationHash` and `label_hash`at the end of the identifier refers to the `featureHash`, so they do not match with each other. For the classification embedding df `featureHash` should be used.

Gorkem-Encord (approved):

PR #454, opened 2023-06-05
[Draft] Partial fix for hash mapping logic, need to discuss what the desired behaviour is before merging.

james-encord:
Fix project view after pressing upload to encord button. (Need to discuss expected behaviour)


james-encord:
Points to discuss:
 - remapping editor url (not loaded from data using the mapping file), check if this is remapped when copy occurs.
 - shifting local -> remote, change keys in database, yes / no?
 - where else needs to actually use the remapped state, and how much we should or should not change.

frederik-encord:
Let's discuss this on the standup

PR #456, opened 2023-06-06
fix: show when explorer is loading data 

frederik-encord:
instead of showing no results message


frontend_components/src/encord_active_components/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -120,6 +124,27 @@ export const Explorer = ({
     );
   }, [itemSet, sortedItems]);
 
+  useEffect(() => {
```
james-encord, line 127: this can be a useMemo instead of useEffect + useState pair

PR #469, opened 2023-06-07
perf(summary): slight performance improvements on the data summary page

Encord-davids:


james-encord (approved):

src/encord_active/app/data_quality/sub_pages/summary.py
```diff
@@ -1,19 +1,12 @@
 from typing import List
 
-import streamlit as st
-
 from encord_active.app.common.components.metric_summary import (
     render_data_quality_dashboard,
     render_label_quality_dashboard,
-    render_metric_summary,
```
frederik-encord, line 8: Perhaps we should remove the unused function then?

frederik-encord (changes requested):

src/encord_active/app/data_quality/sub_pages/summary.py
Encord-davids: Makes sense!

james-encord (approved):

frederik-encord (approved):

PR #472, opened 2023-06-08
fix: obey filters from the toolbox when exporting to coco

frederik-encord:
Fixes two things:
1. The filters applied in the toolbox will now be obeyed - on the label row level
2. Export is much faster because we no longer download data to a directory the we delete right after. One consequence is that if the project is synced with Annotate, the `file_name` property of the images won't be a url to a location on disc but just the file name. Users will have to do the mapping them self. 

For completeness, local projects (not synced with Annotate) will get image entries looking like this:
image
while remote projects synced with Annotate will get
image

### DIFF
--- a/src/encord_active/lib/coco/encoder.py
+++ b/src/encord_active/lib/coco/encoder.py
@@ -257,7 +257,7 @@ class CocoEncoder:
         data_hash = data_unit["data_hash"]
         self._data_hash_to_image_id_map[(data_hash, 0)] = image_id
         return {
-            "coco_url": data_unit["data_link"],
+            "coco_url": "",
             "flickr_url": "",
             "id": image_id,
             "image_title": data_unit["data_title"],
@@ -291,7 +291,7 @@ class CocoEncoder:
             self.download_video_images(url, destination_path, data_hash, video_title)
 
         images = []
-        coco_url = data_unit["data_link"]
+        coco_url = ""
         height = data_unit["height"]
         width = data_unit["width"]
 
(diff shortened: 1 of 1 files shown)


src/encord_active/lib/coco/encoder.py
```diff
@@ -260,29 +261,21 @@ def get_image(self, label_hash: str, data_unit: dict) -> dict:
...
-        )
-        absolute_destination_path = self._download_file_path.joinpath(relative_destination_path)
-        download_condition = self._download_files and not absolute_destination_path.exists()
-
-        if download_condition or self._force_download:
-            self.download_image(url, absolute_destination_path)
-
-        return str(relative_destination_path)
+    def get_local_path_or_file_name(self, data_unit: dict) -> str:
+        url: str = data_unit["data_link"]
```
james-encord, line 273: The signed url in data_link has a risk of being out of date: using project_data_structure.label_structure(label_hash).iter_data_units(data_unit_hash=du_hash).signed_url

this will return the latest cached value from the caching infra, which should be more likely to be up to date. (though for long running streamlit instances this can also be out-of-date - though coco export would be the only part that this effects).

frederik-encord: I just removed the link. That will remove confusion.

src/encord_active/lib/coco/encoder.py
```diff
@@ -256,33 +257,25 @@ def get_image(self, label_hash: str, data_unit: dict) -> dict:
...
```
james-encord, line 273: nit: double check this for local files: the database row is the source of truth for filepaths for the program.

james-encord (approved):

PR #474, opened 2023-06-08
perf: removes uneeded rerendering of 2d embeddings and adds orjson for faster serialization

Encord-davids:


frontend_components/src/encord_active_components/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -116,6 +120,7 @@ export const Explorer = ({
   }, [metrics?.length]);
 
   useEffect(() => {
+    if (isLoadingSortedItems) return;
     setSortedAndFiltered(
       sortedItems?.filter(({ id }) => itemSet.has(id)) || []
     );
```
james-encord, line 126: nit: isLoadingSortedItems should be added to the dependency list (due to correlation with sortedItems not full needed)

2nd nit: is setSortedAndFiltered another memo in disguise?

james-encord (dismissed):

james-encord:
aside: may be worth using webstorm or adding a npm extension to the workspace as there are some changes that should probably show up as lints.

frontend_components/src/encord_active_components/frontend/src/components/explorer/Explorer.tsx
Encord-davids: first nit: Absolutely

2nd nit: not really since we set it from other places as well

Encord-davids:
> aside: may be worth using webstorm or adding a npm extension to the workspace as there are some changes that should probably show up as lints.

we definitely need to setup linting and type checking for the FE and run it on GH as part of our existing/new workflow.

james-encord (approved):

PR #476, opened 2023-06-12
fix: remote subset

james-encord:
Fixes subset execution for remote projects (tested against internal all-in-one project)

Verified upload still works with these changes.

Reduced from earlier version of the pr as some of the code changes were not needed on further investigation.

### DIFF
--- a/src/encord_active/app/actions_page/export_filter.py
+++ b/src/encord_active/app/actions_page/export_filter.py
@@ -457,7 +457,10 @@ def render_subset_button(
                     dataset_description=cols.dataset.description if cols.dataset else None,
                 )
             except Exception as e:
-                st.error(str(e))
+                print(f"EXCEPTION: {e}")
+                import traceback
+                traceback.print_exc()
+                st.error(f"EXCEPTION: {e}")
                 return
 
         set_updates(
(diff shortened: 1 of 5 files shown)


src/encord_active/app/actions_page/export_filter.py
```diff
@@ -482,7 +482,7 @@ def render_export_button(
     disabled_help = None
     if not action_utils:
         disabled_help = "Contact Encord to enable integration"
-    elif not is_filtered:
+    elif is_filtered:
```
james-encord, line 485: discovered when verifying export works. Filter was switched, export to encord button only worked when a filter was selected. This is different to the error message.

src/encord_active/lib/encord/actions.py
```diff
@@ -333,15 +336,26 @@ def create_project(
...
 
+        # Reverse ordering : correct data unit hash mapping data
+        data_hash_mapping = DataHashMapping()
+        for new_du, old_du in dataset_creation_result.new_du_hash_to_original_mapping.items():
+            data_hash_mapping.set(old_du, new_du)
+
         replace_uids(
             self.project_file_structure,
-            dataset_creation_result.lr_du_mapping,
+            dataset_creation_result.lr_du_mapping,  # FIXME: needs to be updated to work correctly for non-images
```
james-encord, line 346: i don't think we support image group creation for non-encord projects so i think this is fine.

frederik-encord (approved):

PR #480, opened 2023-06-13
feat: tag nested classifications

frederik-encord:
Tagging nested classifications at import.
Still doesn't work for project refreshes though.

### DIFF
--- a/src/encord_active/cli/utils/encord.py
+++ b/src/encord_active/cli/utils/encord.py
@@ -94,7 +94,7 @@ Check that you have the correct ssh key set up and available projects on [blue]h
         "project_hash": project.project_hash,
         "ssh_key_path": ssh_key_path.as_posix(),
         "has_remote": True,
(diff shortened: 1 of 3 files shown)


src/encord_active/cli/utils/encord.py
```diff
@@ -62,12 +85,16 @@ def import_encord_project(ssh_key_path: Path, target: Path, encord_project_hash:
...
+    ontology = OntologyStructure.from_dict(project.ontology)
+    option_hashes_to_tag = suggest_tagging_data(ontology)
+
     meta_data = {
         "project_title": project.title,
         "project_description": project.description,
         "project_hash": project.project_hash,
         "ssh_key_path": ssh_key_path.as_posix(),
         "has_remote": True,
+        "nested_attributes_to_tag": list(option_hashes_to_tag),
```
Encord-davids, line 97: [curious]🤔 

Why do we need this as part of the project meta? (we don't seem to use it afterwards)

PR #481, opened 2023-06-13
fix: resolve inconsistency between lock and toml

eloy-encord:


pyproject.toml
```diff
@@ -78,7 +78,7 @@ statsmodels = "^0.13.5"
 umap-learn = "^0.5.3"
 streamlit-plotly-events = "^0.0.6"
 llvmlite = "^0.39.1"                                     # Pinning, as lower versions conflict with other libs
-encord-active-components = "^0.0.22"
+encord-active-components = "0.0.22"
```
james-encord, line 81: https://python-poetry.org/docs/dependency-specification/

use "==0.0.22" for strict exact

james-encord (approved):

pyproject.toml
eloy-encord: that's when using the command `add`

eloy-encord: not what appears in the toml

eloy-encord: btw, you shouldn't make modifications to the toml directly, just use the poetry CLI commands which guarantee that the state is preserved. Direct modification is what got us to this fix.

PR #488, opened 2023-06-14
misc(cli): show absolute paths of output files

eloy-encord:


src/encord_active/cli/print.py
```diff
@@ -80,8 +82,9 @@ def print_ontology(
     json_ontology = json.dumps({"objects": objects, "classifications": classifications}, indent=2)
 
     if state.get("json_output"):
-        Path("./ontology_output.json").write_text(json_ontology, encoding="utf-8")
-        rich.print("Stored mapping in [blue]`./ontology_output.json`")
+        output_file_path = Path("./ontology_output.json").resolve()
```
frederik-encord, line 85:
```suggestion
        output_file_path = Path("./ontology_output.json").expanduser.absolute()
```

frederik-encord (changes requested):

src/encord_active/cli/print.py
```diff
@@ -33,7 +33,9 @@ def print_encord_projects(
     project_query = None if query is None else ProjectQuery(title_like=query)
     json_projects = get_projects_json(app_config.get_or_query_ssh_key(), project_query)
     if state.get("json_output"):
-        Path("./encord-projects.json").write_text(json_projects, encoding="utf-8")
+        output_file_path = Path("./encord-projects.json")
```
james-encord, line 36: no resolve() here?

also the sequence: path.expanduser().absolute().resolve() is the best bet to have the project output a canonical path.

src/encord_active/cli/print.py
james-encord: .expanduser().absolute().resolve() is the best bet for full canonical path

eloy-encord: we can skip `expanduser` as the path is relative only and there is no `~` which expands the user and there is no need of `absolute` when you use `resolve` already.

src/encord_active/cli/print.py
```diff
@@ -113,8 +116,9 @@ def print_data_mapping(
     json_mapping = json.dumps(mapping, indent=2)
 
     if state.get("json_output"):
-        Path("./data_mapping.json").write_text(json_mapping, encoding="utf-8")
-        rich.print("Stored mapping in [blue]`./data_mapping.json`")
+        output_file_path = Path("./data_mapping.json").resolve()
```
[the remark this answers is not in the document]
eloy-encord: same as above

src/encord_active/cli/print.py
eloy-encord: good catch with the resolve

frederik-encord (approved):

PR #491, opened 2023-06-15
fix: prefer relative path uri in database

james-encord:
fixes issue caused by coco files occasionally generated un-normalised uri's.

also means that moving projects works more reliably.

### DIFF
--- a/src/encord_active/lib/common/data_utils.py
+++ b/src/encord_active/lib/common/data_utils.py
@@ -148,7 +148,7 @@ def download_file(
 
 
 def url_to_file_path(url: str, project_dir: Path) -> Optional[Path]:
-    if url.startswith("/") or url.startswith("./"):
+    if url.startswith(("/", "./", "../", "~/")):
         return Path(url)
     if url.startswith("file:"):
         return Path(unquote(urlparse(url).path))
@@ -162,12 +162,12 @@ def url_to_file_path(url: str, project_dir: Path) -> Optional[Path]:
 
 
 def file_path_to_url(path: Path, project_dir: Path, relative: Optional[bool] = None) -> str:
-    path = path.absolute().resolve()
+    path = path.expanduser().absolute().resolve()
     if relative is not None and not relative:
         return path.as_uri()
 
     # Attempt to create relative path
-    root_path = project_dir.absolute().resolve()
+    root_path = project_dir.expanduser().absolute().resolve()
     try:
         rel_path = path.relative_to(root_path)
         return f"relative://{rel_path.as_posix()}"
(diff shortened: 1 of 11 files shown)


src/encord_active/lib/common/data_utils.py
```diff
@@ -146,10 +147,42 @@ def download_file(
...
+        relative_path = url[len("relative://"):]
+        return project_dir / Path(relative_path)
+    if url.startswith("absolute://"):
+        absolute_path = url[len("absolute:/"):]
+        return Path(absolute_path)
+    return None
+
+
+def file_path_to_url(path: Path, project_dir: Path, relative: Optional[bool] = None) -> str:
+    path = path.absolute().resolve()
```
Encord-davids, line 165: If `path` starts with `~/` this statement will yield a bad results, we should use `expanduser` instead of `absolute`.
image

```suggestion
    path = path.expanduser().resolve()
```

src/encord_active/lib/common/data_utils.py
```diff
@@ -146,10 +147,42 @@ def download_file(
...
+    root_path = project_dir.absolute().resolve()
+    try:
+        rel_path = path.relative_to(root_path)
+        return f"relative://{rel_path.as_posix()}"
+    except ValueError:
+        if relative is not None:
+            raise
+
+    # Fallback to strict uri
+    return path.as_uri()
```
frederik-encord: Wouldn't it be simpler to change `relative` to `absolute` to get rid of the `None`?
Or did I miss something?

```suggestion
def file_path_to_url(path: Path, project_dir: Path, absolute: bool = False) -> str:
    path = path.expanduser().resolve()
    if absolute:
        return path.as_uri()

    # Attempt to create relative path
    root_path = project_dir.absolute().resolve()
    try:
        rel_path = path.relative_to(root_path)
        return f"relative://{rel_path.as_posix()}"
    except ValueError:
        raise
```

src/encord_active/lib/common/data_utils.py
```diff
@@ -146,10 +147,42 @@ def download_file(
...
+
+
+def download_image(url: str, project_dir: Path) -> Image.Image:
+    path = url_to_file_path(url, project_dir)
+    if path is not None:
```
frederik-encord, line 184: would the path always exist?

Encord-davids (approved):

src/encord_active/lib/common/data_utils.py
james-encord: no - the path could be remote (or for remote encord projects a signed url)

src/encord_active/lib/common/data_utils.py
james-encord: its used in the exception to allow fallback to absolute or not.

PR #506, opened 2023-06-20
docs: change file names to match expected content and url generation

eloy-encord:


docs/docs/user-guide/exporting.md
```diff
@@ -57,12 +57,12 @@ This format ensures compatibility with various computer vision and machine learn
...
+## Export to Encord {#export-to-encord}
 
 Encord Active also provides the option to export your local project directly to the Encord platform.
 This allows you to seamlessly transfer your data to Encord, where you can leverage additional features and capabilities for annotation, collaboration, and project management.
 
-**Steps to export you local project to Encord Annotate:**
+**Steps to export you local project to Encord:**
 1. Navigate to the explorer pages and use the [standard filter feature][standard-filter-feature] in the toolbox to select the specific data you want to export.
 2. In the same toolbox, locate and access the _Action_ tab.
    ![toolbox-action-tab-local-project](../images/user-guide/toolbox-action-tab-local-project.png)
```
Gorkem-Encord, line 68: I am not sure if we should remove the **Annoate** here. Users, who first used the Encord Active, without the knowledge of Encord Annoate might be confused. Since they (Encord Active, Encord Annotate) are different projects now, we may want to specifically indicate Encord Annotate here. Would like to get feedback from @nikolaj-encord and @david-babuschkin-encord at this point.

david-babuschkin-encord: Personally, I would change it to 'Encord's annotation platform' to be explicit and make it obvious.

Also, it should be "export your local project", there's an 'r' missing in the current text :)

eloy-encord: yeah, Encord is not only for annotation and these past few days with the different products available have made that clearer. I guess Encord platform should suffice.

Gorkem-Encord (approved):

PR #511, opened 2023-06-22
fix: Change default server ports

james-encord:
Currently a lot of users are experiencing issues where they already have some program running on port 8000 causing the explorer page to stop working.

The will require a new encord-active-components version. (will update PR with published new version once accepted)


rad-cord (approved):
Is there a way to have some port-scanning functionality a-la node? Where it tries a few ports in a range if the one it's currently trying is taken? Alternatively, bind to port 0 and propagate that. But both approaches would probably need some run-time introspection so this is good for a quick win 🤷

eloy-encord (changes requested):
fix the checks before merging

james-encord:
> fix the checks before merging

Check failure was ephemeral poetry install issue, seems to have fixed itself.

Encord-davids (approved):

eloy-encord (approved):

PR #514, opened 2023-06-22
fix(coco): support index 0 category

Encord-davids:


src/encord_active/lib/coco/importer.py
```diff
@@ -215,6 +215,8 @@ def create_ontology(self) -> LocalOntology:
             shapes = category_info.shapes
             for i, shape in enumerate(shapes):
                 new_id = cat.id_ * 10 + i
+                if new_id < 10:
+                    new_id = f"0{new_id}"
```
james-encord, line 219: Can we use an integer value instead of adding a special case 0 -> a string?

Encord-davids: Won't make any difference since it is serialized to JSON it will always end up as a string anyway.

But still a fair point for the sake of consistency.

james-encord (dismissed):

PR #518, opened 2023-06-26
feat: add workflow stage as metadata type filter

eloy-encord:


src/encord_active/lib/common/filtering.py
```diff
@@ -33,15 +38,28 @@ def __hash__(self):
...
 
+    # Add workflow stage column in workflow projects
+    if is_workflow_project(pfs):
+        filtered["label_hash"] = filtered.index.str.split("_", n=1).str[0]
+        lr_metadata = json.loads(pfs.label_row_meta.read_text(encoding="utf-8"))
+        lr_to_workflow_stage = {
+            lr_hash: metadata.get("workflow_graph_node", dict()).get("title", None)
+            for lr_hash, metadata in lr_metadata.items()
+        }
+        filtered["workflow_stage"] = filtered["label_hash"].map(lr_to_workflow_stage)
```
Encord-davids, line 54: [nitpick] 💅 

I think adding the column isn't necessary.

```suggestion
        label_hashes = filtered.index.str.split("_", n=1).str[0]
        lr_metadata = json.loads(pfs.label_row_meta.read_text(encoding="utf-8"))
        lr_to_workflow_stage = {
            lr_hash: metadata.get("workflow_graph_node", dict()).get("title", None)
            for lr_hash, metadata in lr_metadata.items()
        }
        filtered["workflow_stage"] = label_hashes.map(lr_to_workflow_stage)
```

src/encord_active/lib/common/filtering.py
```diff
@@ -55,7 +73,10 @@ def apply_filters(df: pd.DataFrame, filters: Filters):
     for column, text_filter in filters.text.items():
         filtered = filtered[filtered[column].astype(str).str.contains(text_filter)]
 
-    filtered.drop(columns=["data_row_id", "is_label_metric"], inplace=True)
+    columns_to_delete = ["data_row_id", "is_label_metric"]
+    if "label_hash" in filtered.columns:
+        columns_to_delete.append("label_hash")
+    filtered.drop(columns=columns_to_delete, inplace=True)
```
Encord-davids, line 79: if the suggestion above works this will not be necessary

```suggestion
    filtered.drop(columns=["data_row_id", "is_label_metric"], inplace=True)
```

src/encord_active/lib/common/filtering.py
```diff
@@ -33,15 +38,28 @@ def __hash__(self):
         return hash(json.dumps(self.dict(exclude_defaults=True)))
 
 
-def apply_filters(df: pd.DataFrame, filters: Filters):
+def apply_filters(df: pd.DataFrame, filters: Filters, pfs: ProjectFileStructure):
     filtered = df.copy()
     filtered["data_row_id"] = filtered.index.str.split("_", n=3).str[0:3].str.join("_")
     filtered["is_label_metric"] = filtered.index.str.split("_", n=3).str.len() > 3
 
+    # Add workflow stage column in workflow projects
```
Encord-davids, line 46: [curious]🤔 

Correct me if I'm wrong, but it seems that we are adding the `workflow_stage` column solely for the purpose of filtering, right?
If that is the case, I think we can move this part of the logic into `filter_workflow_stages` and just not add the column.

eloy-encord: Sure, can do, either way we are already putting this piece of content in the wrong place as the filters should only filter 😄.

eloy-encord: with this change they are more clustered so easier to change, but not great nevertheless

src/encord_active/lib/common/filtering.py
```diff
@@ -102,6 +107,20 @@ def filter_object_classes(to_filter: pd.DataFrame, classes: List[str]):
...
+    label_hashes = to_filter.index.str.split("_", n=1).str[0]
+    lr_metadata = json.loads(pfs.label_row_meta.read_text(encoding="utf-8"))
+    lr_to_workflow_stage = {
+        lr_hash: metadata.get("workflow_graph_node", dict()).get("title", None)
+        for lr_hash, metadata in lr_metadata.items()
+    }
+    to_filter["workflow_stage"] = label_hashes.map(lr_to_workflow_stage)
+
+    filtered_user_input = to_filter[to_filter["workflow_stage"].isin(stages)]
+    return filtered_user_input
```
Encord-davids, line 121: [nitpick] 💅 

Since we don't need the workflow stage after the filter we can avoid adding it to the table.

```suggestion
    filtered_label_hashes = {
        lr_hash
        for lr_hash, metadata in lr_metadata.items()
        if metadata.get("workflow_graph_node", dict()).get("title", None) in stages
    }

    return to_filter[label_hashes.isin(filtered_label_hashes)]
```

Encord-davids (approved):

PR #519, opened 2023-06-27
feat (iterator): iterate on non-labeled data option

Gorkem-Encord:


src/encord_active/lib/common/iterator.py
```diff
@@ -85,6 +86,13 @@ def iterate(self, desc: str = "") -> Generator[Tuple[dict, Optional[Image.Image]
                     self.num_frames = len(label_row.data_units)
                     data_units = sorted(label_row.data_units.values(), key=lambda du: int(du["data_sequence"]))
                     for data_unit in data_units:
+
+                        if self._on_unlabeled_data:
```
james-encord, line 90: should this be _skip_unlabeled_data. 

on_unlabeled_data=True implies not skipping values. (or add not and flip the condition).

Gorkem-Encord: you mean skip_labeled data? because we are skipping those and iterating on unlabeled one.

james-encord (approved):

PR #528, opened 2023-06-28
feat(explorer): convert predictions explorer to react

Encord-davids:


src/encord_active/app/page_menu.py
```diff
@@ -29,7 +29,7 @@
     "Data Drift": project_similarity(),
 }
 
-DEFAULT_PAGE_PATH = ["Data Quality", "Summary"]
+DEFAULT_PAGE_PATH = ["Model Quality", "Explorer"]
```
frederik-encord, line 32: Think this was for testing efficiency purposes no?

src/encord_active/lib/model_predictions/map_mar.py
```diff
@@ -170,7 +170,7 @@ def compute_mAP_and_mAR(
         full_list_to_class_level_idx: Dict[int, int] = {v.item(): i for i, v in enumerate(class_level_to_full_list_idx)}
 
         _ious = ious[pred_select]
-        TP_candidates = set(class_level_to_full_list_idx[_ious >= iou_threshold].astype(int).tolist())
+        TP_candidates = set(class_level_to_full_list_idx[_ious >= (iou_threshold or 0)].astype(int).tolist())
```
frederik-encord, line 173: Not sure why 0 is the choice here? 0 means that everything will be considered, even predictions that don't overlap with any labels. I would at least add a small epsilon I think.

src/encord_active/server/routers/project.py
```diff
@@ -56,10 +75,32 @@ def read_item_ids(
...
+        column = [col for col in merged_metrics.columns if col.lower() == sort_by_metric.lower()][0]
+        df = merged_metrics
+    except:
+        try:
+            if filters.prediction_filters is None:
+                raise
+            df, _ = get_model_predictions(project, filters.prediction_filters)
+
+            if filters.prediction_filters.outcome == ObjectDetectionOutcomeType.FALSE_NEGATIVES:
+                sort_by_metric = re.sub(r"(.*?\()P(\))", r"\1O\2", sort_by_metric)
```
frederik-encord, line 89: Is this really `\10` always?
I understand it as getting a metric like "Object Area (P)" and then returning the same thing, but may be missing something?

src/encord_active/server/routers/project.py
```diff
@@ -88,7 +129,13 @@ def tagged_items(project: ProjectFileStructureDep):
...
-    lr_hash, du_hash, frame, *_ = id.split("_")
+    lr_hash, du_hash, frame, *object_hash = id.split("_")
+
+    row = get_model_prediction_by_id(project, id)
+
+    if row:
+        return to_item(row, project, lr_hash, du_hash, frame, object_hash[0] if len(object_hash) else None)
+
     with DBConnection(project) as conn:
         row = MergedMetrics(conn).get_row(id).dropna(axis=1).to_dict("records")[0]
```
frederik-encord, line 140: Perhaps we should add a fallback if there is nothing in this result. This is where we saw the app crashing previously.

frederik-encord (changes requested):

src/encord_active/app/page_menu.py
Encord-davids: Correct, removed

pyproject.toml
```diff
@@ -92,6 +92,7 @@ ftfy = "^6.1.1"
 regex = "^2023.5.5"
 poethepoet = "^0.16.5"
 orjson = "^3.9.0"
+pyinstrument = "^4.5.0"
```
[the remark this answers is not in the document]
Encord-davids: It is actually quite useful to identify performance issues' root causes. I'm tempted to keep it as a dev dependency, wdyt?

src/encord_active/lib/model_predictions/map_mar.py
Encord-davids: I reverted it back to using the default we had before this. Now the UI will always show the IOU slider.

src/encord_active/server/routers/project.py
Encord-davids: This should work for all false negative metrics, it just replaces the `P` with an `O`. I'll replace the regex with a simpler `str.replace` which will make this part more explicit.

src/encord_active/server/routers/project.py
Encord-davids: We should probably raise an informative error but the app should still fail.
The flow is "if there is a prediction with such id return it otherwise return a row from the merged metrics table".
If the id requested wasn't found anywhere we should probably return a 404.

src/encord_active/server/routers/project.py
frederik-encord: Hahaha my bad 🤦🏻

pyproject.toml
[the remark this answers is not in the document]
frederik-encord: Yes, as dev it's fine 👍🏻

src/encord_active/lib/model_predictions/map_mar.py
frederik-encord: Not sure I 100% understand the implications then?

Encord-davids: Basically, the UI will show the IOU slider with the original default and it will affect the color or the predictions. On top of that, you can filter to only see F/TP or FN.

frederik-encord (dismissed):

frederik-encord (approved):

PR #532, opened 2023-07-03
fix: limit boundaries

Gorkem-Encord:


src/encord_active/cli/utils/coco.py
```diff
@@ -50,7 +50,7 @@ def import_coco_predictions(
         elif res.bbox:
             format, shape = Format.BOUNDING_BOX, Shape.BOUNDING_BOX
             x, y, w, h = res.bbox
-            data = BoundingBox(x=x / img_w, y=y / img_h, w=w / img_w, h=h / img_h)
+            data = BoundingBox(x=x / img_w, y=y / img_h, w=max(1.0, w / img_w), h=max(1.0, h / img_h))
```
james-encord, line 53: bounding box clamping should be.
x = clamp(x)
w = clamp(w + x) - x
similar for y.

but accepting as it is still mostly correct

james-encord (dismissed):

src/encord_active/cli/utils/coco.py
Gorkem-Encord: In addition, the usage of the `max` was also wrong. Fixed it.

james-encord (approved):

PR #536, opened 2023-07-04
fix: only embed on migration if ontology has entries

frederik-encord:


Gorkem-Encord (approved):

eloy-encord (changes requested):
Hi, blocking the approval because of the failed checks, hopefully it's an intermittent one but better safe than sorry.

eloy-encord (approved):

PR #545, opened 2023-07-05
docs: sdk revision

eloy-encord:
Include the revision, refactor and/or remake of subsections contained within the SDK section.
Should be approved after PR #539, so we can be 100% sure that the links to the notebooks added in that PR are in fact correct.


docs/docs/faq.mdx
```diff
@@ -74,8 +74,8 @@ While we're working on improving the efficiency, there are a couple of tricks th
    This will allow you to continuously see what have been computed so far.
    Refresh the browser once in a while when new metrics are done computing in your first terminal.
 
-2. You can also kill the import process as soon as metrics have started to compute. This will leave you with a project containing fewer metrics. As a consequence, you will not be able to see as many insights as if the process is allowed to finish.
-   However, you can always use the [`encord-active metric run`](./cli#run) command to run metrics that are missing.
+2. You can also kill the import process as soon as metrics have started to compute. This will leave you with a project containing less metrics. As a consequence, you will not be able to see as many insights as if the process is allowed to finish.
```
Gorkem-Encord, line 77:
```suggestion
2. You can also kill the import process as soon as metrics have started to compute. This will leave you with a project containing fewer metrics. As a consequence, you will not be able to see as many insights as if the process is allowed to finish.
```

Gorkem-Encord, line 77:
```suggestion
2. You can also kill the import process as soon as metrics have started to compute. This will leave you with a project containing fewer quality metrics. As a consequence, you will not be able to see as many insights as if the process is allowed to finish.
```

Gorkem-Encord (commented):
One thing I am not sure about is having very similar content both in the docs and jupyter notebook (for example for initialize-a-project.mdx we already have two notebooks that have the same content). Because if there is a change in our codebase/SDK, we need to change both pages (jupyter notebook and docs page) separately which will create extra work, we will forget the other one and it will create inconsistency etc. Since we can already write comments/explanations into a jupyter notebook I think we can just keep that one and give reference to that from the docs. Wdyt @david-babuschkin-encord @nikolaj-encord?

PR #550, opened 2023-07-13
feat: standardize importers

eloy-encord:


frederik-encord:
@eloy-encord this branch is out of date. Let's update it and review. 
I also just wanted to check that the docs will align with these changes.

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,504 @@
...
-            "/path/to/data_root/some_name.png"
+    :param project: The project to import the predictions into.
+    :param predictions: The predictions in Encord's Prediction class format to be imported.
+    """
+    if len(predictions) == 0:
+        return
 
-    :param du_hash_name_lookup: A transform ``f(file_path) -> (data_unit_hash: str, frame: int)`` to translate between
-        data units and their associated mask file names. You can omit this argument if you
+    with PredictionWriter(project) as writer:
```
frederik-encord, line 74: The importer will break if there are different types of predictions in the same import. I would add a check up front here and check two things.

1. Assert that all predictions are either classifications or objects.
2. If they are all objects, ensure that the `prediction.format` is consistent across all predictions

If that's not the case throw a meaningful error such that users know what the limitations / issues are

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,504 @@
...
-def import_mask_predictions(
-    project: EncordProject,
-    data_root: Path,
-    cache_dir: Path,
-    prediction_writer: PredictionWriter,
-    class_map: Optional[Dict[str, int]] = None,
-    file_name_regex: str = PNG_FILE_NAME_REGEX,
-    du_hash_name_lookup: Callable[[Path], Tuple[str, int]] = None,
-):
+def import_predictions(project: Project, predictions: list[Prediction]):
```
frederik-encord, line 64: Love this simplicity!

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,504 @@
...
+    # Retrieve the mapping from ontology object hashes to object class ids
+    if ontology_mapping is None:
+        ontology_mapping = _json_load(predictions_json.parent / "ontology_mapping.json")
 
-    # === Prepare ontology lookup === #
-    with (data_root / "ontology_label_map.json").open("r", encoding="utf-8") as f:
-        label_name_map = json.load(f)
+    # Retrieve the mapping from COCO image ids to data unit hashes
+    if image_mapping is None:
+        image_mapping = _json_load(predictions_json.parent / "image_mapping.json")
```
frederik-encord, line 247: `__json_load` raises an error. Should we perhaps handle that?

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,504 @@
...
-        o["name"]: o["featureNodeHash"] for o in project.ontology["objects"] if o["shape"] in BoxShapes
+    # Verify the validity of the ontology object hashes, ensuring they exist and represent bounding boxes
+    pfs = ProjectFileStructure(project_dir)
+    relevant_ontology_hashes = {
+        o.feature_node_hash: o.shape for o in OntologyStructure.from_dict(json.loads(pfs.ontology.read_text())).objects
     }
-    hash_lookup: Dict[str, Optional[str]] = {v: object_name_to_hash.get(k) for k, v in label_name_map.items()}
+    bad_hashes = [h for h in ontology_mapping.keys() if h not in relevant_ontology_hashes]
+    if len(bad_hashes) > 0:
+        raise ValueError(f"The ontology mapping contains references to invalid ontology object hashes: {bad_hashes}")
```
frederik-encord, line 256: Let's tell what the options are in this error message.

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,504 @@
...
+    # Verify the validity of the ontology object hashes, ensuring they exist and represent bounding boxes
+    pfs = ProjectFileStructure(project_dir)
+    relevant_ontology_hashes = {
+        o.feature_node_hash
+        for o in OntologyStructure.from_dict(json.loads(pfs.ontology.read_text())).objects
+        if o.shape.value in BoxShapes
+    }
+    bad_hashes = [h for h in ontology_mapping.keys() if h not in relevant_ontology_hashes]
+    if len(bad_hashes) > 0:
+        raise ValueError(f"The ontology mapping contains references to invalid ontology object hashes: {bad_hashes}")
```
frederik-encord, line 361: Let's tell what the options are.

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,504 @@
...
+    # Verify the validity of the ontology object hashes, ensuring they exist and represent polygons
+    pfs = ProjectFileStructure(project_dir)
+    relevant_ontology_hashes = {
+        o.feature_node_hash
+        for o in OntologyStructure.from_dict(json.loads(pfs.ontology.read_text())).objects
+        if o.shape.value == ObjectShape.POLYGON
+    }
+    bad_hashes = [h for h in ontology_mapping.keys() if h not in relevant_ontology_hashes]
+    if len(bad_hashes) > 0:
+        raise ValueError(f"The ontology mapping contains references to invalid ontology object hashes: {bad_hashes}")
```
frederik-encord, line 457: Also give options here.

frederik-encord (changes requested):

src/encord_active/lib/model_predictions/importers.py
eloy-encord: On point 2, seems like the original COCO predictions importer sent predictions with two object formats, `polygon` and `bbox`, so probably we should continue allowing the same behaviour.

src/encord_active/lib/model_predictions/importers.py
eloy-encord: Probably the best approach is to document the exception, as there is no recovery workaround if the files don't exist.

src/encord_active/lib/model_predictions/importers.py
```diff
@@ -55,214 +57,522 @@
...
+    relevant_ontology_hashes = {
+        o.feature_node_hash
+        for o in OntologyStructure.from_dict(json.loads(pfs.ontology.read_text())).objects
+        if o.shape.value == ObjectShape.POLYGON
+    }
+    bad_hashes = [h for h in ontology_mapping.keys() if h not in relevant_ontology_hashes]
+    if len(bad_hashes) > 0:
+        raise ValueError(
+            f"The ontology mapping contains references to invalid ontology object hashes: {bad_hashes}.\n"
+            "Please, update the mapping with the correct object hashes from the project's ontology."
```
frederik-encord, line 474: What I meant by these was printing the `relevant_ontology_hashes` directly. 🤷🏻

frederik-encord (approved):

PR #553, opened 2023-07-14
fix: move react query network mode to always.

james-encord:
Otherwise this means that the explorer page will fail to load in offline conditions, as the server is local this should not prevent connections to local ea instances when offline.


james-encord:
Accept & i'll update the PR to contain a new version of encord_active_components before merging.

frontend_components/src/encord_active_components/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -77,15 +78,19 @@ export const Explorer = ({
 
   const api = getApi(projectName, authToken, baseUrl);
 
+  console.log('api', api);
   const { data: hasPremiumFeatures } = useQuery(
     ["hasPremiumFeatures"],
-    api.fetchHasPremiumFeatures,
-    { staleTime: Infinity }
+      () => {
+          console.log('fetch premium request', );
```
frederik-encord, line 85: This seems to have slipped
```suggestion
```

frederik-encord (changes requested):

frederik-encord (approved):

frederik-encord (approved):

PR #559, opened 2023-07-26
misc: improve messaging in metrics

eloy-encord:


frederik-encord (approved):
Approved this although we have brand new metric definitions coming very soon.

PR #560, opened 2023-07-27
fix: update to new docs url

Gorkem-Encord:


src/encord_active/app/app_config.py
```diff
@@ -56,7 +56,7 @@ def get_or_query_ssh_key(self) -> Path:
 Encord Active needs to know the path to the [blue]private ssh key[/blue] which is associated with Encord.
 Don't know this? Please see our documentation on the topic to get more help.
 
-[blue]https://docs.encord.com/docs/annotate-public-keys#set-up-public-key-authentication[/blue]
+[yellow]https://docs.encord.com/docs/annotate-public-keys[/yellow]
```
eloy-encord, line 59: why to change the url?

eloy-encord: isn't already pointing to the relevant section?

Gorkem-Encord: Strange, this page could not open so I thought due to the latest update of the docs, copied the new url without looking  at the ling but they are the same :D

PR #564, opened 2023-08-01
fix: docker build

Encord-davids:


src/encord_active/server/start_server.py
```diff
@@ -7,7 +7,9 @@ def start(path: Path, reload=False):
     from uvicorn import run
 
     environ["SERVER_START_PATH"] = path.as_posix()
-    run("encord_active.server.main:app", reload=reload, port=8000)
+    opts = {"reload": reload, "port": environ.get("PORT"), "host": environ.get("HOST")}
```
james-encord, line 10: nit: i'd make the default. values explicit in the .get for PORT & HOST instead of relying on uvicorn behaviour.

james-encord (approved):

src/encord_active/server/start_server.py
Encord-davids: I don't think we should override the defaults unless explicitly done by the user.
If on the next version of `uvicorn` they decide (for a good reason) to change the default we won't get the change.
(overriding defaults with the same defaults is usually a bad idea)

PR #571, opened 2023-08-03
chore: better build setup

Encord-davids:


frontend_components/encord_active_components/frontend/src/constants.ts
```diff
@@ -1,2 +1,2 @@
-export const apiUrl = import.meta.env.VITE_API_URL ?? "http://localhost:8000";
-export const env = import.meta.env.VITE_ENV || "local";
+export const apiUrl = "http://localhost:8000";
```
james-encord, line 1: can we make this similarly just a env variable? & extend the vite.config.ts logic to always set the value.

Encord-davids: Went with a slightly different solution but fixed the issue

james-encord (dismissed):

james-encord (approved):

james-encord (approved):

james-encord (approved):

PR #574, opened 2023-08-04
[Very-WIP][Do Not Merge] Remove disk dependency

james-encord:
Leaving this open to track the changes in the no-disk branch before it is ready to merge.


src/encord_active/analysis/executor.py
```diff
@@ -0,0 +1,1177 @@
...
+                            data_distances_list.append(data_raw_similarities[data_raw_index])
+                        data_indices = np.stack(data_indices_list)
+                        data_distances = np.stack(data_distances_list)
+                    else:
+                        metrics_tqdm.set_description(f"{metrics_tqdm_desc}  [Creating index]")
+                        nn_descent = NNDescent(
+                            data=data,
+                            parallel_batch_queries=True,
+                            n_jobs=-1,
+                        )
```
Gorkem-Encord, line 833: Shall we set metric='cosine', so that it is bounded in [-1, 1]?

src/encord_active/analysis/types.py
```diff
@@ -0,0 +1,135 @@
...
+    points: Optional[PointTensor]
+    mask: Optional[MaskTensor]
+    bounding_box: Optional[BoundingBoxTensor]
+
+
+@dataclass(frozen=True)
+class NearestNeighbors:
+    metric_deps: list[MetricDependencies]
+    metric_keys: list[Union[Tuple[uuid.UUID, int], Tuple[uuid.UUID, int, str]]]
+    similarities: list[float]
```
Gorkem-Encord, line 34: What if we change `similarities` to `distances`, which is more self-explanatory?

src/encord_active/analysis/metrics/active_learning/uniqueness.py
```diff
@@ -0,0 +1,32 @@
...
+            ident="metric_image_uniqueness",
+            long_name="Image Uniqueness",
+            desc="How unique an image is in the dataset",
+            metric_type=MetricType.NORMAL,
+        )
+
+    def calculate(self, deps: MetricDependencies, annotation: Optional[AnnotationMetadata]) -> MetricResult:
+        if annotation is not None:
+            return None
+        # FIXME: validate normalisation of this value (is 1.0 distance clamp reasonable??)
```
Gorkem-Encord, line 25: Yeah, with the cosine distance, it is reasonable.

src/encord_active/analysis/executor.py
```diff
@@ -0,0 +1,1177 @@
...
+                            data_hash=du_meta.data_hash,
+                            du_hash=du_meta.du_hash,
+                        ),
+                        du_meta=du_meta,
+                    )
+                    for meta, du_meta in grouped_batch
+                ]
+
+                batch_inputs = BaseFrameBatchInput(
+                    images=torch.stack([frame.image for frame in frame_inputs]),
```
Gorkem-Encord, line 471: Images may have different sizes, so the stack operation here fails in that case right?

Gorkem-Encord:
Generic question: is `calculate_batched()` only suitable for img_group or vidoe type data units?

src/encord_active/analysis/metrics/image/area.py
```diff
@@ -0,0 +1,85 @@
...
+            metric_type=MetricType.UINT,
+        )
+
+    def calculate(
+        self, deps: MetricDependencies, image: ImageTensor, mask: Optional[MaskTensor], bb: Optional[BoundingBoxTensor]
+    ) -> MetricResult:
+        if mask is None:
+            return float(image_width(image)) * float(image_height(image))
+        else:
+            return cast(torch.Tensor, deps["ephemeral_grayscale_image"]).shape[0]
```
Gorkem-Encord, line 39: Can we just remove `cast` from all over the code as it seems a bit dangerous, instead what about checking the type with `isinstance()`, like:

```
if isinstance(deps["ephemeral_grayscale_image"], torch.Tensor):
     return deps["ephemeral_grayscale_image"].shape[0]
 else:
    raise TypeError(f"Expected torch.Tensor, got {type(deps["ephemeral_grayscale_image"])}")
```

src/encord_active/analysis/metrics/image/contrast.py
```diff
@@ -0,0 +1,57 @@
...
+            desc="",
+            metric_type=MetricType.NORMAL,
+        )
+
+    def calculate(
+        self, deps: MetricDependencies, image: ImageTensor, mask: Optional[MaskTensor], bb: Optional[BoundingBoxTensor]
+    ) -> MetricResult:
+        # Max-std : [0, 255] = 255/2 = 127.5
+        grayscale_image = cast(torch.Tensor, deps["ephemeral_grayscale_image"])
+        return torch.std(grayscale_image.type(torch.float32)).cpu() / 127.5
```
Gorkem-Encord, line 36: We may need to make this change:

```suggestion
        return torch.std(grayscale_image.type(torch.float32), correction=0).cpu() / 127.5
```

according to [here](https://pytorch.org/docs/stable/generated/torch.std.html)

src/encord_active/analysis/metrics/image/sharpness.py
```diff
@@ -0,0 +1,47 @@
...
+            metric_type=MetricType.NORMAL,
+        )
+
+    def calculate(
+        self, deps: MetricDependencies, image: ImageTensor, mask: Optional[MaskTensor], bb: Optional[BoundingBoxTensor]
+    ) -> MetricResult:
+        # Max range of laplacian kernel = [-4, 4] * 255 = [-1020, 1020]
+        # Max value of variance = (2.0 * 1020.0 * 1020.0) = 2080800.
+        laplacian = cast(torch.Tensor, deps["ephemeral_laplacian_image"])
+        return math.sqrt(torch.var(laplacian).cpu().item() / 2080800.0)
```
Gorkem-Encord, line 39: We also need a correction=0 here

src/encord_active/analysis/metrics/image/sharpness.py
```diff
@@ -0,0 +1,47 @@
...
+            long_name="Sharpness",
+            desc="",
+            metric_type=MetricType.NORMAL,
+        )
+
+    def calculate(
+        self, deps: MetricDependencies, image: ImageTensor, mask: Optional[MaskTensor], bb: Optional[BoundingBoxTensor]
+    ) -> MetricResult:
+        # Max range of laplacian kernel = [-4, 4] * 255 = [-1020, 1020]
+        # Max value of variance = (2.0 * 1020.0 * 1020.0) = 2080800.
```
Gorkem-Encord, line 37: Where does this multiplication with 2 come from, I think It should be 1020*1020=1040400

james-encord:
```
>>> torch.var(4 * torch.tensor([-255.0,255.0]))
tensor(2080800.)
```

src/encord_active/analysis/metrics/image/area.py
james-encord: yes - though easiest is probably to use a typed dict for the dependency dict.

src/encord_active/analysis/executor.py
james-encord: added option for different distance metrics

PR #588, opened 2023-08-10
feat: avg simple metrics

Encord-davids:


src/encord_active/db/foo.sqlite
james-encord, line 1: probably not meant to be included

james-encord: maybe add *.sqlite exclusion to gitignore to prevent this.

frederik-encord:
Closing this experimental PR.

PR #593, opened 2023-08-11
fix: prevent sending ids

Encord-davids:


frederik-encord (commented):
Why would we filter identifiers on the active end and not on the premium api?

frederik-encord (approved):

PR #601, opened 2023-08-16
fix: refresh item filters when tags change

frederik-encord:


frederik-encord:
Stal and a better solution on the horizon.

PR #605, opened 2023-08-24
fix: refresh expired urls

frederik-encord:


frederik-encord:
This was not the right solution. Another one is coming up.

PR #606, opened 2023-08-29
perf: tagging

frederik-encord:
Tagging was very slow, when tagging big amounts of data. This PR taks tagging 30K items on my mac from ~30sec to ~3 sec. Not I think that major bottleneck is communitation.

### DIFF
--- a/src/encord_active/server/routers/project.py
+++ b/src/encord_active/server/routers/project.py
@@ -2,7 +2,18 @@ import json
 import shutil
 import uuid
 from enum import Enum
-from typing import Annotated, Dict, List, Optional, Union, cast
+from functools import partial
+from typing import (
+    Annotated,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    TypedDict,
+    Union,
+    cast,
+    overload,
+)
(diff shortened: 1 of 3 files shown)


src/encord_active/server/routers/project.py
```diff
@@ -394,28 +467,72 @@ def _get_or_create_tag_hash(name: str) -> uuid.UUID:
...
+                                (du_hash, frame, annotation_hash), set()
+                            ).difference(new_label_tag_uuids)
+                        ]
+                    )
+                )
+
+        # Delete left over data and annotation tags
+        if data_tags_to_add:
+            sess.execute(
+                text(
```
james-encord, line 496: change this to sess,add(ProjectTaggedX()) - sqlalchemy will automatically batch the changes. May be faster.

frederik-encord:
On my machine, using `sess.insert` is slightly slower (1.77 sec vs. 1.34 - as opposed to 10.9 sec on main). However, the code is cleaner, so sticking with it.

james-encord (dismissed):

james-encord (approved):

PR #612, opened 2023-09-01
fix: normalize distance to border metric

Gorkem-Encord:


src/encord_active/analysis/metrics/annotation/distance_to_border.py
```diff
@@ -21,25 +21,24 @@
 class DistanceToBorderMetric(OneObjectMetric):
     def __init__(self) -> None:
         super().__init__(
-            ident="metric_border_relative",  # FIXME: rename ident to something better??
-            long_name="Distance in pixels to the border",
-            desc="",
+            ident="metric_distance_to_border_relative",
```
james-encord, line 24: keep the ident small the same - changing it breaks db logic as its the column key. (& needs to be kept small due to maximum name limit).

james-encord (approved):

PR #633, opened 2023-09-07
fix: add missing argument

frederik-encord:


src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -933,7 +933,11 @@ const GalleryItem = ({
   onShowSimilar: JSX.IntrinsicElements["button"]["onClick"];
   iou?: number;
 }) => {
-  const { data, isLoading } = useApi().fetchItem(itemId, iou);
+  const { data, isLoading } = useApi().fetchItem(
+    itemId,
+    iou !== undefined,
```
Encord-davids, line 938: We should check the scope instead, I think `iou` can be undefined

Encord-davids (changes requested):

src/encord_active/frontend/src/components/explorer/Explorer.tsx
frederik-encord: @Encord-davids I fixed this now could you give it a look?

Encord-davids (approved):

PR #654, opened 2023-09-28
WIP: 🚀 Initial Layout Redesign

deoracord:
Initial Layout redesign 
1. Changing layout to match Figma files [here](https://www.figma.com/file/wX94ndsPuQp0EsbVF7O4Ih/Active-Improvements?type=design&node-id=879-29135&mode=design&t=Tu6FuuoEzUv7PNoA-0) 
2. Adding new tabs for overview, filters and display.


src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -405,276 +428,296 @@ export function Explorer({
...
+                      premiumSearchState.setSearch(args);
+                    },
+                  }}
+                />
+              ),
+              right: (
+                <Segmented
+                  selected
+                  value={metricDomain}
+                  options={["Data", "Label"]}
```
Encord-davids, line 460: two thing:

* We should store the options somewhere that it is reusable and typed
* separating label and value is usually the better approach, e.g.
```tsx
<Segment options={[{ label: "Data", value: "data"/index }]}  />
```

src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -405,276 +428,296 @@ export function Explorer({
...
+                  onChange={
+                    handleSelectedDomainChange as (val: SegmentedValue) => void
```
Encord-davids, line 462: Let's avoid casting types, if typescript says the type is wrong it probably is 😄

src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -405,276 +428,296 @@ export function Explorer({
...
+                  </span>
+                ),
+                key: "gridView",
+                children: (
+                  <div className="flex h-full flex-col items-center bg-gray-100 py-2">
+                    <Space
+                      style={{
+                        flex: "0 1 0",
+                        height: "100%",
+                      }}
```
Encord-davids, line 488: [nitpick] 💅 

I think this could be done with classname

src/encord_active/frontend/src/components/explorer/Explorer.tsx
james-encord: for data vs annotation. AnalysisDomain is the openapi type i use in most places.

src/encord_active/frontend/src/components/explorer/display/Display.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <div className=" text-sm text-gray-8">Confidence</div>
+        <Slider
+          range
+          defaultValue={[0, 1]}
+          min={0}
+          max={1}
+          step={0.01}
+          className="w-full"
+          onAfterChange={(newRange: [number, number]) => {
+            setConfidenceFilter(
```
james-encord, line 79: May want to keep confidence as just a standard metric filter for now as we don't support it in the TP/FP/FN condition - so special case logic may confuse the user.

Also - this can get out of sync with a confidence filter added manually to the standard list of metric filters (though i think that errors).

src/encord_active/frontend/src/components/preview/AnnotatedImage.tsx
```diff
@@ -99,7 +99,7 @@ export function AnnotatedImage(props: {
       ) : (
         <img
           ref={image}
-          className="!rounded-none object-contain transition-opacity"
```
james-encord, line 102: I think that this may be the cause of the de-sync between annotations and images when being displayed. The svg positioning & size metadata needs to be checked.

src/encord_active/frontend/src/components/explorer/display/Display.tsx
```diff
@@ -0,0 +1,132 @@
...
+      </div>
+      <div className="flex items-center justify-between">
+        <div className="px-2 py-4 text-sm text-gray-8">Display Predictions</div>
+        <Switch disabled className="bg-gray-9" />
+      </div>
+      <div className="flex flex-col items-start justify-between p-4">
+        <div className=" text-sm text-gray-8">Confidence</div>
+        <Slider
+          range
+          defaultValue={[0, 1]}
```
james-encord, line 73: use of defaultValue with a source of truth state elsewhere can cause de-sync between values. (i.e. this filter displayed 1 range, the other filter a different range). As the updates are not directly correlated.

src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -131,6 +146,48 @@ export function Explorer({
...
+  const [selectedMetricData, setSelectedMetricData] = useState<Metric>({
+    domain: "data",
+    metric_key: "metric_random",
+  });
+  const [selectedMetricLabel, setSelectedMetricLabel] = useState<Metric>({
+    domain: "annotation",
+    metric_key: "metric_random",
+  });
+
+  const handleSelectedDomainChange = (val: AnalysisDomain) => {
```
james-encord, line 173: selectedMetric could be a derived value (doesn't even need memo).

const selctedMetric = analysisDomain == "data ? selectedMetricData. : selectedMetricLabel

src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -131,6 +146,48 @@ export function Explorer({
...
+    setAnalysisDomain(val);
+    if (val === "data") {
+      setSelectedMetric(selectedMetricData);
+    } else if (val === "annotation") {
+      setSelectedMetric(selectedMetricLabel);
+    }
+  };
+
+  const handleMetricChange = (val: Metric) => {
+    setSelectedMetric(val);
```
james-encord, line 183: similarity this can just update one of the 2 states.

src/encord_active/frontend/src/components/explorer/display/Display.tsx
deoracord: Fixed, Made confidence bounds and values populate from FIlterState for annotation filters.

src/encord_active/frontend/src/components/explorer/Explorer.tsx
```diff
@@ -131,14 +168,30 @@ export function Explorer({
     setShowAnnotations(selectedMetric.domain === "annotation");
   }, [selectedMetric.domain, setShowAnnotations]);
 
+  // Data or Label selection
+  const analysisDomainOptions = useMemo(
```
james-encord, line 172: This could be a global constant (use strict typing for completeness instead of generation via a mapping).

src/encord_active/frontend/src/components/explorer/ExplorerSearchResults.tsx
```diff
@@ -96,41 +96,47 @@ function ExplorerSearchResultsRaw(props: {
...
+            setSimilaritySearch={setSimilaritySearch}
+            selected={selectedItems === "ALL" || selectedItems.has(item)}
+            hideExtraAnnotations={showAnnotations}
+            featureHashMap={featureHashMap}
+            iou={iou}
+            gridCount={gridCount}
+          />
+        )}
+      />
+      <div className="hidden h-full">1</div>
```
Encord-davids, line 139: [curious]🤔 

What is the intention behind this?

src/encord_active/frontend/src/components/explorer/ExplorerTypes.ts
```diff
@@ -1,5 +1,14 @@
...
+  AnalysisDomain,
+  PredictionDomain,
+  SearchFilters,
+} from "../../openapi/api";
 
+export type Metric = {
+  domain: "data" | "annotation";
+  metric_key: string;
+};
+export type MetricDomain = "data" | "label";
```
Encord-davids, line 11: This seems unused

```suggestion
```

src/encord_active/frontend/src/components/explorer/ExplorerSearchResults.tsx
deoracord: Removed, was added when I was trying to give the parent full height of the remaining page and make it overflow.

PR #656, opened 2023-10-04
Fix (import project): remove forbidden chars in file name

Gorkem-Encord:


Gorkem-Encord:
removed due to [this PR](https://github.com/encord-team/encord-active/pull/665), which covers the current PR.

PR #661, opened 2023-10-06
Feat: no-disk -> fix image uniqueness

Gorkem-Encord:
similarity value, which is bounded between [-1, 1], is converted to distance value by normalizing between [0, 1]

### DIFF
--- a/src/encord_active/analysis/metrics/active_learning/uniqueness.py
+++ b/src/encord_active/analysis/metrics/active_learning/uniqueness.py
@@ -29,4 +29,4 @@ class ImageUniqueness(DerivedMetric):
         else:
             # Clamp large distances to (unique)
             # FIXME: filter out nearby frames / same video?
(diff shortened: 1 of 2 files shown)


james-encord:
checked annotation quality as well or is the scaling not necessary. (Negative values may cause issues in what i remember of the formula as well).

Gorkem-Encord:
> checked annotation quality as well or is the scaling not necessary. (Negative values may cause issues in what i remember of the formula as well).

There is no problem in annotation quality, it was solved in one of the previous PR.

Gorkem-Encord:
> checked annotation quality as well or is the scaling not necessary. (Negative values may cause issues in what i remember of the formula as well).

Need a fix there also, good catch.

james-encord:
I think this should be closed: https://github.com/encord-team/encord-active/pull/667 should fix everything.

PR #666, opened 2023-10-11
Feat/init collections

deoracord:


src/encord_active/db/migrations/versions/2023_10_10_173753-add_timestamps_to_tags.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+import sqlalchemy as sa
+import encord_active
+import sqlmodel
+from alembic import op
+from sqlalchemy import Text
+
+
+# revision identifiers, used by Alembic.
+revision = "e33776b0b984"
```
james-encord, line 16: change revision to same format as others: 20231010173753

src/encord_active/db/alembic.ini
```diff
@@ -59,8 +59,8 @@ version_path_separator = os  # Use os.pathsep. Default configuration used for ne
 output_encoding = utf-8
 
 # Path to the database used as the source of truth for migrations
-# sqlalchemy.url = sqlite:///../../../local_tests/encord-active.sqlite
-sqlalchemy.url = postgresql://postgres@localhost:5433/encord-active
+sqlalchemy.url = sqlite:///../../../local_tests/encord-active.sqlite
+# sqlalchemy.url = postgresql://postgres@localhost:5433/encord-active
```
james-encord, line 63: Undo this. PG gen gives much more accurate schema changes.

src/encord_active/db/models/project_tag.py
```diff
@@ -12,6 +12,8 @@ class ProjectTag(SQLModel, table=True):
     project_hash: UUID = field_uuid()
     name: str = field_text()
     description: str = field_text()
+    created_at: datetime = field_datetime(nullable=True)
```
james-encord, line 15: if nullable: Optional[datetime] else nullable=False

src/encord_active/server/routers/project2_tags.py
```diff
@@ -52,15 +62,29 @@ def route_list_tags(project_hash: uuid.UUID, engine: Engine = Depends(dep_engine
         ).fetchall()
     return [ProjectTagEntry(hash=tag_hash, name=name) for tag_hash, name in tags]
 
+@router.get("/meta")
+def route_list_tags(project_hash: uuid.UUID, engine: Engine = Depends(dep_engine_readonly)) -> List[ProjectTagEntryMeta]:
```
james-encord, line 66: tag meta is expensive - so only run against all tags that this is actually needed for. & this is a name clash on the routes.

route should be similar to:

```
@router.get("/{tag_hash}/meta")
 route_tag_meta(project_hash: uuid.UUID, tag_hash: uuid.UUID, engine: Engine = ...) -> TagMetadata:
```

james-encord (changes requested):
Added some comments r.e the api.

src/encord_active/server/routers/project2_tags.py
deoracord: `.where(ProjectTag.project_hash == project_hash)`

It is only running against tags needed. We need to show all tags for a project in the new collections tab.

PR #672, opened 2023-11-27
Feat (DB): small public api

Gorkem-Encord:


src/encord_active/public/active_project.py
```diff
@@ -0,0 +1,77 @@
...
+    def get_ontology(self) -> Dict:
+        active_project = Table("active_project", self._metadata, autoload_with=self._engine)
+
+        stmt = select(active_project.c.project_ontology).where(active_project.c.project_hash == f"{self._project_hash}")
+
+        with self._engine.connect() as connection:
+            result = connection.execute(stmt).fetchone()
+            ontology = result[0]
+
+        return ontology
```
frederik-encord, line 34: Would be better to return an ontology structure?
```suggestion
        return OntologyStructure.from_dict(ontology)
```
Remember the import

frederik-encord (changes requested):

Gorkem-Encord:
Split into two as #673 and #674.

